PR #10, opened 2016-05-11
Tweaking ESLint rules and ES6ing some stuff

pdehaan:
ES6ify a few things and add a couple more ESLint rules and tweaks.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1,11 +1,7 @@
 const assert = require('chai').assert;
-const jsdom = require('../node_modules/jsdom');
+const jsdom = require('jsdom');
 
-const fathom = require('../fathom'),
-      dom = fathom.dom,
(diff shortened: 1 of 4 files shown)


test/test.js
```diff
@@ -1,11 +1,7 @@
 const assert = require('chai').assert;
-const jsdom = require('../node_modules/jsdom');
+const jsdom = require('jsdom');
```
erikrose: You got this to work? I wonder what's wrong with my setup....

erikrose:
Thanks for the idiomaticness; keep it coming! I've been doing 90% Python for a long time, so this really helps.

erikrose:
Looking forward to Travis supporting node 6 so we can get the tests running automatically.

PR #21, opened 2016-06-20
Clusters

erikrose:


utils.js
```diff
@@ -1,4 +1,48 @@
...
+            const key = by(item);
+            if (isBetter(key, bestKeySoFar) || isFirst) {
+                bestSoFar = item;
+                bestKeySoFar = key;
+                isFirst = false;
+            }
+        },
+        iterable);
+    if (isFirst) {
+        throw 'Tried to call best() on empty iterable';
```
pdehaan: ⚠️ Probably an uber-nit, but not sure if we want to make sure we throw exceptions instead of string literals.

Ref: http://eslint.org/docs/rules/no-throw-literal

> _"The fundamental benefit of Error objects is that they automatically keep track of where they were built and originated."_

erikrose: Yes, we should definitely throw Errors. Thanks!

erikrose: Fixed in 88b9e67!

PR #26, opened 2016-06-28
Add glorious CHANGELOG

pdehaan:
The CHANGELOG is pretty verbose (pretty much a 1:1 for each commit).
In other projects we use something like [**conventional-changelog**](https://github.com/conventional-changelog/conventional-changelog-cli), but then that requires using specifically formatted commit messages (see [**mozilla/fxa-content-server** Git Commit Guidelines](https://github.com/mozilla/fxa-content-server/blob/master/CONTRIBUTING.md#git-commit-guidelines)).

But pretty easy to generate using `npm run changelog`. ¯\\\_(ツ)\_/¯

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,149 @@
+1.1.0 / 2016-06-27
+==================
+
+  * Bump version to 1.1.
+    Add version history.
+  * Merge for...const fix for Firefox compatibility.
+  * Remove prefer-const rule due to Firefox for...of issue
+  * Replace for const... with let for Firefox compat
+    This is due to a bug in Firefox, see https://bugzilla.mozilla.org/show_bug.cgi?id=1094995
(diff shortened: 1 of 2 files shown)


pdehaan:
Not sure if there is an easy way to auto-regenerate this whenever you do an npm publish, or if we just do it manually. Or if we even need a full changelog since you outline the major commits in the README doc.

erikrose:
Sorry, somehow I didn't get a mail about this! :-o I'm a big fan of changelogs that highlight the major fixes and features and leave the little stuff to the commit log—it probably stems from a proclivity toward DRY. Do you feel we need a comprehensive one? I do agree we need an easy way for framework-shoppers to see what's new, which is why I generally land version histories in the readme.

erikrose:
Closing this for now so it doesn't just languish; feel free to reopen if you want to keep talking about it.

PR #28, opened 2016-07-09
Add missing comma in README example

rgtk:
Overlookment, it seems.


pdehaan:
@rgtk Good catch, thanks. 🍻

PR #31, opened 2016-07-13
Add coveralls badge to README

pdehaan:
Collect coveralls code coverage info from Travis-CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ language: node_js
 node_js:
   - 6
 
+after_success:
+  - "npm run coveralls"
+
 notifications:
   email:
     on_success: change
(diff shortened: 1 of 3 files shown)


erikrose:
lgtm if tests pass. (Merge at will.)

erikrose:
However, what's that noisy, puzzling comment the coveralls bot just made?

pdehaan:
> However, what's that noisy, puzzling comment the coveralls bot just made?

\*shrug\*
I just disabled a checkbox on the coveralls.io page, so I think it shouldn't add comments for future PRs. But there are a few other flags there if you want to fail builds if the coverage decreases by x% in a PR.

erikrose:
I don't have a strong opinion about it, as long as we have a graph and can tell when it happened (which it looks like we do). If you wanted to put a big number like 10 in there, that would be fine with me. Or just leave it alone.

PR #32, opened 2016-07-14
Iterate properly over querySelectorAll() results even in implementati…

erikrose:
…ons where the results don't conform to the iterator protocol.

NodeLists are guaranteed only to have length and item() props.


jaredkerim:
r+!

PR #40, opened 2016-09-30
Add more ESLint rules and node plugin

pdehaan:
Overall very boring PR.

- Upgraded **eslint** to latest version.
- Added [**eslint-plugin-node**](http://npm.im/eslint-plugin-node) ESLint plugin.
- Added a few new ESLint rules and auto fixed some new errors due to rules changes.
- Switched to function shorthand in a few places.

As always, feel free to take it or leave it.


.eslintrc
```diff
@@ -1,18 +0,0 @@
-env:
```
pdehaan: ESLint now recommends using a file extension for your config file. See http://eslint.org/docs/user-guide/configuring#configuration-file-formats.

So this was renamed to .eslintrc.yml below.

.eslintrc.yml
```diff
@@ -0,0 +1,32 @@
...
+  es6: true
+  mocha: true
+  node: true
+
+extends:
+  - eslint:recommended
+  - plugin:node/recommended
+
+plugins:
+  - node
```
pdehaan: Adds "eslint-plugin-node" and it's recommended rules.

index.js
```diff
@@ -45,9 +45,9 @@ function ruleset(...rules) {
             // digested. Rules run in no particular guaranteed order.
             while (nonterminals.length) {
                 const [inNode, inFlavor] = nonterminals.pop();
-                for (let rule of getDefault(rulesByInputFlavor, inFlavor, () => [])) {
+                for (const rule of getDefault(rulesByInputFlavor, inFlavor, () => [])) {
```
pdehaan: This was all fallout from the new `prefer-const` rule. Since the `rule` value never changes, it didn't like the `let`.
I very vaguely recall that Firefox has/had some issues w/ `for (const foo of bar) {...}`, but I guess it depends on if somebody can use this in a browser or not.
We can also just delete/disable the `prefer-const` rule and pretend none of this ever happened.

utils.js
```diff
@@ -65,7 +65,7 @@ function sum(iterable) {
 
 function length(iterable) {
     let num = 0;
-    for (let item of iterable) {
+    for (const item of iterable) { // eslint-disable-line no-unused-vars
```
pdehaan: Disabled the `no-unused-vars` rule for this line since it was throwing a warning(?) on the unused `item` variable.

erikrose:
I haven't taken a good look at your PR yet, but I thought I'd mention that I'd been working off https://github.com/airbnb/javascript#hoisting and had gotten precisely that far down the page, picking and choosing as I went—in case you want to pick up there.

pdehaan:
We could definitely extend the air-bnb ESLint rules if you want to follow their style guide (we use it in the [Test Pilot](https://github.com/mozilla/testpilot/blob/0f508baf81a8de5d830057182124ccbd7b24ea5c/.eslintrc#L4) repo).
It's pretty robust and well thought out, but I find that it tends to change between revisions, so keeping it up to date without having to tweak code can sometimes be mildly annoying.

erikrose:
At another glance, I see this is against 1.x. I'm very close to finishing 2.0 on the 2.0dev branch, where I've done some additional eslint improvement. Feel free to have a look at that if you like. Otherwise, I'll get to this later and see what we can moosh together.

erikrose:
2.0 is very close to landing now, so I've just picked out the applicable bits and landed them on the 2.0dev branch. Thanks, @pdehaan!

.eslintrc.yml
```diff
@@ -0,0 +1,32 @@
...
+
+root: true
+
+rules:
+  node/exports-style: [error, module.exports]
+  node/no-unpublished-require: [error, {allowModules: [chai, jsdom]}]
```
erikrose: I'm hitting the no-unpublished-require rule with a new package now (`diff`). Do you understand what makes that happen? I don't see anything from https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unpublished-require.md.

PR #53, opened 2017-01-02
Allow running rulesets against elements as well as documents

staktrace:
I'd like to be able to run a fathom ruleset on a subtree of the DOM (i.e. rooted at some element in the DOM, rather than on the entire DOM). Seeing as the only thing fathom needs is a querySelectorAll, which is available on Element as well as Document, it can trivially allow this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,11 +93,14 @@ textContainer fnodes emitted        assign "countedWords" type
 
 Remember that Fathom's rulesets are unordered, so any rule's output can flow into any other rule, not just ones that happen to come lexically after it.
 
-Once the ruleset is defined, run a DOM tree through it:
+Once the ruleset is defined, run a DOM tree or a subtree through it:
 
 ```javascript
+var myDom = jsdom.jsdom("<html><head>...</html>"));
 // Tell the ruleset which DOM to run against, yielding a factbase about the document.
-var facts = rules.against(jsdom.jsdom("<html><head>...</html>"));
+var facts = rules.against(myDom);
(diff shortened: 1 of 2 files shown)


staktrace:
In retrospect this change is probably not a good idea. There's actually no code changes needed to just pass an element into `against` and have it work, and I don't know if this change will break stuff with JS DOM implementations (my use of fathom has been in-browser using browserify). The Travis results seem to imply that it will break.

However, I still think it's a good idea to "officially" support running fathom on an arbitrary element, although that might just be a matter of updating the documentation to say it's supported, and ensure that any functions called on the document object have element-subtree fallbacks.

erikrose:
Like you, I don't know that we need to make any source changes, but I do like the freedom of passing in a subtree. Want to close this and submit a doc patch instead? I patched up the linting situation to address the mostly-spurious indentation errors and re-ran your job, and now it complains only about the Document and Element symbols coming out of nowhere. (Obviously, with zero code changes, there would be zero errors.)

Now that I think of it, a welcome addition with the doc change would be a test or two which passes a subtree in, just to make sure (1) that works and (2) we don't break it in the future.

Thanks for the good thought!

staktrace:
Any particular reason you didn't want to merge the patch?

erikrose:
My mistake! I didn't see that you did, in fact, turn this into a test+doc patch. I'll merge it in.

erikrose:
Thank you for bringing this up and writing such a nice, thorough test for it! :-D

PR #60, opened 2017-03-11
Add .babelrc file so fathom can be used as a dep in webpack / babel p…

jezell:
…rojects

My builds were failing due to missing .babelrc when trying to require fathom.


erikrose:
Thanks!

PR #67, opened 2017-07-07
Headless chrome

swathiiyer2:
able to determine correct image with reasonableish accuracy on very few test cases (need more data); not using annealer; check output with karma debug runner at http://localhost:9876/debug.html


examples/products.js
```diff
@@ -0,0 +1,155 @@
...
+        window.document.body.removeChild(document.getElementById('fixture'));
+    }
+    console.log(stats.score());
+    return stats.score();
+}
+
+/** Return (expected DOM, source DOM) for all the readbaility test docs. */
+function productDocPairs() {
+  return [
+    [staticDom(fs.readFileSync(__dirname + '/../test/product_classification_test_data/amazon/expected-image.html', 'utf-8')),
```
erikrose: It sure is a lot of ugly noise, but I don't know what else you could have done that brfs could have followed.

examples/readability.js
```diff
@@ -15,7 +15,7 @@
  *   over, loosening constraints each time, if it fails
  * * Happily finds body text in things other than divs and p tags.
  */
-const {readFileSync} = require('fs');
+var fs = require('fs');
```
erikrose: Is this because brfs isn't smart enough to recognize readFileSync if we import it deeply?

test/demos.js
```diff
@@ -0,0 +1,112 @@
+const assert = require('chai').assert;
```
erikrose: Before we merge this (or before we merge chrome-headless, if that's where we did the deletes), we should squash or rewrite so we aren't removing most of the test files and then adding them back—it'll mess up our ability to use `git blame` and obscure the few changes we did make to those files.

utils.js
```diff
@@ -398,8 +398,8 @@ function isDomElement(thing) {
  * execute any of its inline scripts.
  */
 function staticDom(html) {
-    const {jsdom} = require('jsdom');
-    return jsdom(html, {features: {ProcessExternalResources: false}});
+    var parser = new DOMParser();
```
erikrose: Ooo, a handy thing! So, I see the `<script>` are still in there. This makes me suspect our test harness actually fetches those with remote sources at runtime. For speed, durability, and the ability to work offline, test harnesses shouldn't hit the network. Is there something we can do about that?

examples/products.js
```diff
@@ -0,0 +1,155 @@
...
+ */
+class DiffStats {
+    constructor(contentFnodes) {
+        this.lengthOfExpectedTexts = 0;
+        this.lengthOfDiffs = 0;
+        this.contentFnodes = contentFnodes || tunedImageFnodes();
+    }
+
+    compare(expectedDom, sourceDom) {
+        //compare images by src
```
swathiiyer2: Had the same thought actually, and recently deleted this and switched to 0/1s!

examples/readability.js
swathiiyer2: Yes, tried it the other way as well, brfs doesn't like it.

utils.js
swathiiyer2: I've read that setting the script types to "application/json" would prevent them from loading. Do you think this is reasonable, or might there be a better way?

examples/products.js
```diff
@@ -0,0 +1,155 @@
...
+
+const leven = require('leven');
+
+const {dom, props, out, rule, ruleset, score, type} = require('../index');
+const {domSort, inlineTextLength, linkDensity, staticDom} = require('../utils');
+const {Annealer} = require('../optimizers');
+
+
+function euclideanDistance(nodeA, nodeB){
+  var rectA = nodeA.getBoundingClientRect();
```
swathiiyer2: alright! might this function make more sense in utils?

utils.js
erikrose: I was hoping there would be a flag in DOMParser or something. Failing all else, we can hand-remove the script tags for now, but we'll eventually want a way to parse untrusted pages server-side and still measure distances.

swathiiyer2: The spec (https://w3c.github.io/DOM-Parsing/) says that the default for the text/html type is for script elements to be marked unexecutable. Does this mean the resource is not fetched at all, or just that the code does not get executed?

erikrose:
Good question, though I'm encouraged to hear that that much is true. Why not try it? You could set up a local server serving the JS and see if it gets hit.

utils.js
swathiiyer2: Alright, I've tried setting up the local server, and it does not get hit, so we've avoided this problem I believe!

erikrose:
Very good news!

examples/products.js
erikrose: Ultimately, once it's good enough to be general-purpose yes.

examples/products.js
```diff
@@ -37,7 +37,7 @@ function euclideanDistance(nodeA, nodeB){
   var rectB = nodeB.getBoundingClientRect();
   var x = rectB.left - rectA.left;
   var y = rectB.bottom - rectB.bottom;
-  return Math.sqrt(x**2 + y**2);
+  return Math.sqrt(Math.pow(x,2) + Math.pow(y,2));
```
erikrose: Curious: why this change?

examples/products.js
```diff
@@ -79,22 +115,28 @@ function tunedImageFnodes() {
       //better score for larger images
       rule(type('images'), score(imageSize)),
 
+      //jpegs used more often than pngs
```
erikrose: A very interesting idea!

examples/products.js
swathiiyer2: It was giving me some sort of syntax error

docpairs.js
```diff
@@ -0,0 +1,201 @@
+var fs = require('fs');
```
erikrose: This shouldn't be at the top level of the fathom package, since it isn't intended for reuse. The `examples` folder is probably the right place. You may want to make a `products` folder there and move all the product-related files within.

docpairs.js
```diff
@@ -0,0 +1,201 @@
...
+const {staticDom} = require('./utils');
+
+function productImageDocPairs() {
+  return [
+    [staticDom(fs.readFileSync(__dirname + '/test/product_classification_test_data/amazon/expected-image.html', 'utf-8')),
+      fs.readFileSync(__dirname + '/test/product_classification_test_data/amazon/source.html', 'utf-8')
```
erikrose: This long line and the ones like it each appear several times. It's bad enough that brfs makes us write the whole thing out a bazillion times; let's minimize it by having a separate function return all the source.html files in a list and then zip them together with the expected data for images, prices, titles, etc.

examples/products.js
```diff
@@ -0,0 +1,155 @@
...
+  return Math.sqrt(x^2 + y^2);
+}
+
+function tunedImageFnodes() {
+    function imageSize(fnode) {
+      return fnode.element.offsetWidth * fnode.element.offsetHeight;
+    }
+
+    function imageHasSrc(fnode){
+      return fnode.element.hasAttribute('src') && fnode.element.getAttribute('src') != '';
```
pdehaan: @erikrose Me thinks the ESLint rule you want is [`eqeqeq`](http://eslint.org/docs/rules/eqeqeq).

examples/products/docpairs.js
```diff
@@ -0,0 +1,111 @@
+var fs = require('fs');
+var zip = require('zip-array').zip
```
erikrose: That shorthand I use sometime works here (using *destructuring assignment*, which you should read up on if you don't know it): `const {zip} = require('zip-array');`

erikrose:
We decided not to go in this direction; headless Chrome or our stuff around it was leaking RAM like crazy.

PR #71, opened 2017-08-16
when() can be called on type() and dom()

swathiiyer2:


lhs.js
```diff
@@ -118,12 +145,11 @@ class DomLhs extends Lhs {
...
-        for (let i = 0; i < matches.length; i++) {  // matches is a NodeList, which doesn't conform to iterator protocol
-            const element = matches[i];
-            ret.push(ruleset.fnodeForElement(element));
-        }
-        return ret;
+        const fnodes = Array.from(matches).map(function (obj) {
+            return ruleset.fnodeForElement(obj);
+        });
+        const predicates = this.predicates;
+        return checkPredicates(predicates, fnodes);
```
erikrose: Why not just say `this.predicates` on this line?

lhs.js
```diff
@@ -15,6 +15,24 @@ function dom(selector) {
     return new DomLhs(selector);
 }
 
+/*
+ * Of all the dom nodes selected by type() or dom(), return only
+ * the fnodes that satisfy all the predicates imposed by calls to
+ * when()
```
erikrose: I was going to suggest that, rather than passing in `predicates`, you just make this a method of Lhs and have it access `this.predicates` itself. Reading the code, I see you've even already documented it that way. :-)

lhs.js
```diff
@@ -15,6 +15,24 @@ function dom(selector) {
...
+ */
+function checkPredicates(predicates, fnodes) {
```
erikrose: I'd rather have this named after the thing it returns. In the Fathom codebase, [as is my usual practice](https://www.grinchcentral.com/function-names-to-verb-or-not-to-verb), I've used imperative names, like "checkFact", to imply no return value and only side effects (like the raising of an exception, in that case). I would suggest `satisfyingWhens` or `fnodesSatisfyingWhens` or something.

lhs.js
```diff
@@ -15,6 +15,24 @@ function dom(selector) {
...
+    const ret = [];
+    fnodes.forEach (function (item) {
```
erikrose: `fnode` would be a better type clue than `item`.

erikrose: weird spacing before paren (here and below)

lhs.js
```diff
@@ -15,6 +15,24 @@ function dom(selector) {
...
+        const result = predicates.every (function (func) {
```
erikrose: `func` → `predicate` (or `pred` if you feel like misreads would be too likely) would make this easier to read through consistency with "predicates". And actually, this would be a good place to use an arrow function, just for brevity.

erikrose: `result` is always a suspiciously uninformative name. I think `matchesPredicates` or `doesMatchPredicates` or `fnodeMatches` or `fnodeMatchesPredicates` would make the `if` below read more clearly.

lhs.js
```diff
@@ -117,13 +144,10 @@ class DomLhs extends Lhs {
...
     fnodes(ruleset) {
-        const matches = ruleset.doc.querySelectorAll(this.selector);
         const ret = [];
-        for (let i = 0; i < matches.length; i++) {  // matches is a NodeList, which doesn't conform to iterator protocol
-            const element = matches[i];
-            ret.push(ruleset.fnodeForElement(element));
-        }
-        return ret;
+        const matches = ruleset.doc.querySelectorAll(this.selector);
+        Array.from(matches).forEach(element => ret.push(ruleset.fnodeForElement(element)));
```
erikrose: There's no need to use Array.from(). NodeLists have their own forEach() method.

lhs.js
```diff
@@ -42,6 +47,28 @@ class Lhs {
...
+    }
+
+    /*
+     * Of all the dom nodes selected by type() or dom(), return only
+     * the fnodes that satisfy all the predicates imposed by calls to
+     * when()
+     */
+    fnodesSatisfyingWhens(fnodes) {
+        const ret = [];
+        const predicates = this._predicates;
```
erikrose: Why the introduction of this new `predicates` symbol?

test/lhs_tests.js
```diff
@@ -51,4 +51,28 @@ describe('LHS', function () {
         const best = facts.get('best');
         assert.equal(best.length, 1);
     });
+
+    it('testing when() on type()', function () {
+        const doc = staticDom('<p id="fat"><p id="fat"></p></p>');
```
erikrose: We ought not nest paragraph tags, even in our tests, as that's invalid HTML.

test/lhs_tests.js
```diff
@@ -51,4 +51,28 @@ describe('LHS', function () {
         const best = facts.get('best');
         assert.equal(best.length, 1);
     });
+
+    it('testing when() on type()', function () {
```
erikrose: This test doesn't discriminate between when() doing its job correctly and it simply returning everything. I recommend having 2 paragraphs of different IDs and checking to make sure it grabs the right one.

And then, of course, there is no need for the max() rule of the score(5).

test/lhs_tests.js
```diff
@@ -51,4 +51,28 @@ describe('LHS', function () {
...
+        const facts = rules.against(doc);
+        const best = facts.get('best');
+        assert.equal(best.length, 2);
+    });
+
+    it('testing when() on dom()', function () {
+        const doc = staticDom('<p id="bat"></p><p></p><p id="fat"></p><p id="bat"></p>');
+        const rules = ruleset(
+            rule(dom('p').when(fnode => fnode.element.id === 'bat'), type('when').score(5)),
+            rule(type('when').max(), out('best'))
```
erikrose: I don't think max() or score(5) add anything to the test. Am I missing something?

lhs.js
```diff
@@ -42,6 +47,28 @@ class Lhs {
         }
     }
 
+    when(pred) {
+        this._predicates.push(pred);
```
erikrose: I'm sorry; I should have noticed this earlier. LHSs are immutable so repetitive ones can be factored out. Thus, this should act like TypeLhs.type() and TypeLhs.max() in returning a fresh LHS.

Also, we should probably take another cue from TypeLhs.type() in that later when()s should override earlier ones. This lets us factor out repetition. (Here's [the documentation about similar behavior in the RHS](http://mozilla.github.io/fathom/ruleset.html?highlight=repetition#right-hand-sides).) So the _predicates array can be replaced with a simple value. If we someday want to be able to combine when()s, we could do it through, perhaps, and `andWhen()` method. Or things might change a whole lot before then; it's best not to worry about it unless we need it immediately.

lhs.js
swathiiyer2: I think the forEach switches the context, so it can't access this._predicates.

lhs.js
swathiiyer2: For some reason, it's not able to find that method. I'm not sure what the issue could be, I'm just using matches.forEach(blah), and it gives me "TypeError: matches.forEach is not a function".

erikrose: jsdom must not have implemented it. Makes me want to go back to the `for` loop.

lhs.js
erikrose: Ah, you're right. It changes `this` to point to your anonymous function. In those cases, it's idiomatic to say something like `const self = this` outside the function to preserve the old `this`. Your choice is actually superior, because it doesn't preserve anything you don't need.

lhs.js
```diff
@@ -32,7 +32,7 @@ function dom(selector) {
  */
 class Lhs {
     constructor() {
-        this._predicates = [];
+        this._predicate = function(fnode){return true;};
```
erikrose: A shorter spelling would be `() => true`.

lhs.js
```diff
@@ -145,7 +145,9 @@ class DomLhs extends Lhs {
     fnodes(ruleset) {
         const ret = [];
         const matches = ruleset.doc.querySelectorAll(this.selector);
-        Array.from(matches).forEach(element => ret.push(ruleset.fnodeForElement(element)));
+        for(let i = 0; i < matches.length; i++) {
```
erikrose: space after keywords, e.g. `for`

lhs.js
```diff
@@ -48,8 +48,10 @@ class Lhs {
     }
 
     when(pred) {
-        this._predicate = pred;
-        return this;
+        const selector = this._type !== undefined? this._type : this.selector;
```
erikrose: So Lhs doesn't have to know about its subclasses (which is a bad pattern in general because it's a circular dependency: you can never make a new subclass without modifying the superclass), why not introduce a clone() method that TypeLhs and DomLhs can implement and Lhs can specify? Otherwise, this looks good as far as immutability goes.

lhs.js
```diff
@@ -144,10 +141,16 @@ class DomLhs extends Lhs {
         this.selector = selector;
     }
 
+    clone(pred) {
+        let lhs = new this.constructor(this.selector);
+        lhs._predicate = pred;
```
erikrose: Since this is called simply "clone", it shouldn't also set a predicate. You could rename it "cloneWithPredicate" or something, but I recommend simply moving the predicate-setting out of it and into the caller, since a clone method is often a useful primitive and will likely prove useful in the future.

docs/ruleset.rst
```diff
@@ -43,6 +43,14 @@ Left-hand sides are currently a few special forms which select nodes to be fed t
 
 .. autofunction:: and(typeCall[, typeCall, ...])
 
+.. function :: when(predicate)
+
+    Further specify types of node you'd like to select. Can be chained with type() or dom().
```
erikrose: Can you move this into a jsdoc comment in the code itself, like the other functions are?

erikrose: Also, you can do the fancy formatting on type() and dom() so they link to the docs for those routines. I believe it's `:func:type()`, probably with a bit more to disambiguate it from RHS's type().

docs/ruleset.rst
```diff
@@ -43,6 +43,14 @@ Left-hand sides are currently a few special forms which select nodes to be fed t
...
+
+    Example: dom('p').when(fnode => fnode.element.id === 'foo')
```
erikrose: And you can stick double-backticks around the code to make it show up monospaced in the docs.

swathiiyer2:
Think I've fixed the formatting, let me know if there's anything else though, I'm not too familiar with ReST!

lhs.js
```diff
@@ -42,6 +47,33 @@ class Lhs {
...
+     */
+    fnodesSatisfyingWhen(fnodes) {
+        const ret = [];
+        const pred = this._predicate;
+        fnodes.forEach(function (fnode) {
+            if (pred(fnode)) {
+                ret.push(fnode);
+            }
+        });
+        return ret;
```
mythmon: Is there a reason not to shorten this to a call to filter, like below?

```js
fnodesSatisfyingWhen(fnodes) {
    return fnodes.filter(this._predicate);
}
```

lhs.js
```diff
@@ -116,14 +148,17 @@ class DomLhs extends Lhs {
...
+    }
+
     fnodes(ruleset) {
-        const matches = ruleset.doc.querySelectorAll(this.selector);
         const ret = [];
-        for (let i = 0; i < matches.length; i++) {  // matches is a NodeList, which doesn't conform to iterator protocol
-            const element = matches[i];
-            ret.push(ruleset.fnodeForElement(element));
+        const matches = ruleset.doc.querySelectorAll(this.selector);
+        for (let i = 0; i < matches.length; i++) {
```
mythmon: You could wrap `matches` in a call to `Array.from`, so you can use the iterator protocol:

```js
fnodes(ruleset) {
    const matches = ruleset.doc.querySelectorAll(this.selector);
    const ret = Array.from(matches).map(match => ruleset.fnodeForElement(match));
    return super.fnodesSatisfyingWhen(ret);
}

Comment (swathiiyer2) [reply lhs.js]: I did originally have it using Array.from, but Erik preferred the for loop!

File: lhs.js (hunk 18 above)
Comment (swathiiyer2) [reply lhs.js]: Ah yes forgot about filter, will change this!

File: lhs.js (hunk 19 above)
Comment (erikrose) [reply lhs.js]: (I was trying to avoid copying the whole thing.)

File: lhs.js [hunk 20]
```diff
@@ -64,14 +64,7 @@ class Lhs {
...
     fnodesSatisfyingWhen(fnodes) {
-        const ret = [];
-        const pred = this._predicate;
-        fnodes.forEach(function (fnode) {
-            if (pred(fnode)) {
-                ret.push(fnode);
-            }
-        });
-        return ret;
+        return Array.from(fnodes).filter(this._predicate);
```
erikrose: I believe `fnodes` is always an array already, so you should be able to just call `filter` on it.

docs/ruleset.rst
```diff
@@ -45,11 +45,6 @@ Left-hand sides are currently a few special forms which select nodes to be fed t
 
 .. function :: when(predicate)
```
erikrose: The when() docs don't show up for me when I do the build (go into `docs`, `make html`, then look at `_build/html/index.html` in a browser). I notice there's a space before your ::. I don't know if ReST is that picky, but it wouldn't surprise me. Ah, and I see your when() comment starts with `/*`. It needs to start with `/**`. It's a jsdoc thing.

lhs.js
swathiiyer2: Hm, so I tried that but it didn't work. When I printed fnodes, in one instance it was an array, but in the other it was a set.

erikrose:
Ah, that could be. But in that case, I'm giving a wrong-typed "default" argument to getDefault elsewhere in this file (`() => []`).

swathiiyer2:
I searched the project for anywhere this might be happening, and I think it's either in rule.js (line 201) or fnode.js (line 117). Should I just leave this as is? Also, were you able to see when() in the docs?

lhs.js
erikrose: (Moving this back on-thread. Apparently, replies to GH emails no longer go to the right thread.) https://github.com/mozilla/fathom/blob/0b8153831d93456762357268b6e573dd66c128d7/lhs.js#L159 was the one I meant. It technically is correct, since [all I guarantee from fnodes()](https://github.com/mozilla/fathom/blob/0b8153831d93456762357268b6e573dd66c128d7/lhs.js#L54) is an iterable, but it's asking for trouble to be inconsistent.

erikrose: But whether you choose to correct that or not, all I guarantee is an iterable, so you would have to do Array.from or use a more general filter() procedure that isn't a method of Array. There's probably one in wu.

swathiiyer2: Alright! So I should just leave this as is? Also, were you able to see when() in the docs?

mythmon: Since Erik is on PTO this week, I'll step in. This can probably be left as is.

I rendered the docs, and I see `when` in them, but without any details, just as a bare function reference. I'm not sure how to fix that. Can you render the docs locally, and see if you get the same thing?

swathiiyer2: I actually hadn't written any details in, Erik told me to move them to a jsdoc comment, but maybe I interpreted incorrectly and I was supposed to leave them in the docs as well. I'm having a lot of trouble installing sphinx to render the docs. I can add the details in, but I'm not sure what to try now to get that fixed?

mythmon: Sphinx should be pulling the details from the jsdoc comments, so you shouldn't need anything in the .rst file. To render the docs, you can create a python virtualenv, and then install the requirements from [docs/tooling/doc-building-requirements.txt](https://github.com/mozilla/fathom/blob/master/docs/tooling/doc-building-requirements.txt) into it. Then you also need to `npm install jsdoc`. Finally, you can run `make html` in the doc directory.

If you have trouble with that, lets chat on IRC to see if we can get it sorted out.

swathiiyer2: Ok there we go, the docs should be correct now

PR #72, opened 2017-08-31
attributes

swathiiyer2:


test/utils_tests.js
```diff
@@ -37,4 +37,76 @@ describe('Utils', function () {
                           'The graph has a cycle.');
         });
     });
+
+    describe('searchAttributes', function () {
```
mythmon: Would it make sense to have a negative test, that doesn't find anything? Is that a concept that makes sense here?

utils.js
```diff
@@ -391,6 +391,29 @@ function isDomElement(thing) {
...
+ * @arg attrs {string} the attributes you want to search. If none provided, default is to search all.
+ * @return true if any of the attributes satisfy some condition
+ */
+function searchAttributes(fnode, searchFunction, ...attrs) {
+    // if attrs argument not provided, default is to search all attributes
+    const attributes = attrs.length === 0? Array.from(fnode.element.attributes).map(a => a.name) : attrs;
+    for (let i = 0; i < attributes.length; i++) {
+        const attr = fnode.element.getAttribute(attributes[i]);
+        // If the attribute is an array, apply the scoring function to each element
+        if (attr && ((attr.isArray && attr.some(searchFunction)) || searchFunction(attr))) {
```
mythmon: In what case would `attr` be falsey? Could that be something that searchers may want to be able to search for?

swathiiyer2: When a user specifies some attribute they want to check, but the node doesn't actually have that attribute, attr would be falsey. I think searchers could check this easily just using element.hasAttribute before they put it into this function, I just wrote this as a precaution.

test/utils_tests.js
swathiiyer2: Do you think its sufficient to assume the negative works given that the other tests show that it chooses the correct one out of two? (it finds that one node passes the check while the other doesn't?)

utils.js
mythmon: I think it would be valuable to have a test that doesn't find anything.

Is it possible for the search to return multiple things? That would also be a useful test if so.

swathiiyer2:
Anymore changes needed?

test/utils_tests.js
```diff
@@ -108,5 +108,72 @@ describe('Utils', function () {
...
+            function scoreFunc(fnode) {
+                return searchAttributes(fnode, searchFunc)? 5 : 1;
+            }
+
+            function searchFunc(attr) {
+                return attr.includes('z');
+            }
+            const facts = rules.against(doc);
+            const best = facts.get('best');
+            assert.equal(best.length, 2);
```
mythmon: Is this `2` because nothing matches and therefore are all "the best" in that they are all equal?

test/utils_tests.js
```diff
@@ -108,5 +108,72 @@ describe('Utils', function () {
             assert.equal(best[0].element.id, 'fat');
         });
 
+        it('test non-attribute', function () {
```
mythmon: I don't understand this test title. How is this "non-attribute"? It still seems to be using attributes, in the form of `rule(type('attr'), score(scoreFunc))`. Does "attribute" means something special in Fathom?

test/utils_tests.js
swathiiyer2: yes!

test/utils_tests.js
swathiiyer2: Mm I was referring to a html element's attributes (e.g. id, class, etc.) rather than the type I've given it in fathom, and was trying to test the case when a user tries to search an attribute that the element does not have (non-attribute?).

mythmon: Oh, I see. The test here is that the first element has the attribute, and the second one doesn't, so the second doesn't get included in the results? A comment to that affect could be useful.

test/utils_tests.js
```diff
@@ -37,4 +37,143 @@ describe('Utils', function () {
...
+            }
+            const facts = rules.against(doc);
+            const best = facts.get('best');
+            assert.equal(best.length, 1);
+            assert.equal(best[0].element.id, 'sat');
+        });
+
+        it('search array attribute', function () {
+            const doc = staticDom(`
+                <img id = "fat" class="fat bat sat" ></img><img id ="foo" class="foo bar boo"></img>
```
mythmon: As a small nit, the spacing around the equals signs in these tests is inconsistent. It would be good to be consistent about that.

erikrose:
Again (since I'm not sure you saw it on IRC), thank you, @mythmon, for reviewing while I was away. I rested easy knowing Swathi would have good feedback in my absence. :-)

PR #76, opened 2017-10-19
Add rollup config for creating a single-file build of Fathom.

Osmose:
I wanted to use fathom in a webextension, but didn't want to bother setting up webpack or another module bundler. So I threw together a quick [Rollup](https://rollupjs.org/) config to build a single `fathom.js` file using UMD. When dropped in a webpage, it should add a `fathom` global with all the exports from the library on it. After installing the dev dependencies, you can run it with `./node_modules/.bin/rollup -c`.

I made a minor attempt at seeing if I could run the test suite against it, and failed; mocha seems to error out with the message: `Cannot read property 'regeneratorRuntime' of undefined`, referring to a reference to `global.regeneratorRuntime`. I think it has something to do with how [wu.js](https://github.com/fitzgen/wu.js/) is Webpacked, but I'm not really sure.

I tested this using the example from the [Basic Use docs](https://mozilla.github.io/fathom/using.html) and it seems to work. 

I doubt this is ready for merging and needs work to be reliable and usable, but I figured sharing it was better than not.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /*.log
 /coverage
 /docs/_build
+/dist
--- a/index.js
+++ b/index.js
@@ -7,11 +7,14 @@ const {ruleset} = require('./ruleset');
 const {dom} = require('./lhs');
 const {out} = require('./rhs');
 const {and, atMost, conserveScore, max, note, props, score, type, typeIn} = require('./side');
+const utils = require('./utils');
+const clusters = require('./clusters');
 
 
 module.exports = {
     and,
     atMost,
+    clusters,
     conserveScore,
     dom,
     max,
@@ -22,5 +25,6 @@ module.exports = {
     ruleset,
     score,
     type,
-    typeIn
+    typeIn,
+    utils,
 };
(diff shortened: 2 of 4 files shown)


erikrose:
Great! I would want to get the tests running against it to prove it worked, but that might be made easier by the news that I'm looking to break our dependency on wu. wu isn't maintained anymore, and we really don't use much from it. So I think our order of operations here is to copy and paste the 3 routines we use from wu into Fathom and then see if rollup passes the tests any better.

Osmose:
One problem I just found with this is that the util functions aren't exposed in the main module, and thus aren't available in the rollup build. Simplest fix would be to export utils in `index.js`.

erikrose:
Exporting utils sounds fine to me. I didn't have a strong reason for not doing it, and roller-uppers weren't on my mind at that point.

erikrose:
Want to add it to the PR?

Osmose:
Added. I also added `clusters` for the same reason.

erikrose:
Thanks! Do the tests pass against it now (and how are you running them, anyway?), or should we mark the rollup stuff as experimental in the docs?

Osmose:
I only attempted to run them by manually changing the import paths in a test file. Didn't really try much after that.

Given that it seems to work in my one tiny webextension, marking this as experimental in the docs and adding an issue for hooking up testing seems like the best way forward.

erikrose:
Thanks again!

PR #77, opened 2018-06-14
Fix npm install after cloning the repo

motin:
Updated package-lock.json (previous one resulted in 404 Not Found: ansi-styles@2.2.0)


erikrose:
Thanks, motin!

PR #81, opened 2018-08-17
Land ES6 module-ization

erikrose:


utilsForBackend.mjs
```diff
@@ -0,0 +1,27 @@
...
+ */
+function staticDom(html) {
+    return jsdom(html, {features: {ProcessExternalResources: false,
+                                   FetchExternalResources: false}});
+}
+
+module.exports = {
+    dirsIn,
+    staticDom
+};
```
pdehaan: Instead of `module.exports = {...}` should we be doing something like `export function dirsIn(path) {...}`?

erikrose:
You are absolutely right. This one slipped through!

clusters.mjs
```diff
@@ -321,7 +321,7 @@ function clusters(fnodes, splittingDistance, getDistance = distance) {
     return matrix.clusters();
 }
 
-module.exports = {
+export {
```
pdehaan: Personal curiosity, but is there any difference between using `export { clusters, distance };` versus using `export function distance() {...}` and `export function clusters() {...}`?

I'm seeing both styles in this PR, but didn't know if there was a specific reason or pros/cons to different styles.

erikrose: No excuse, sir! I think the inconsistency comes from my flapping around trying to get rollup to work. There were infinite permutations to test, and I finally schunked into something that worked and considered myself lucky. I will go back and homogenize. I prefer the `export function` spelling because it doesn't require double-entry bookkeeping.

Issue #97, opened 2019-06-28
Consider adding a `fathom-pick list` command or similar to get the list of sample filenames


biancadanforth:
[FathomFox](https://github.com/mozilla/fathom-fox)'s Vectorizer can generate a `vectors.json` file for multiple sample pages at once, though it requires the sample pages to be entered as a multiline string where each line is a sample page's filename, e.g.:
```
0001 Walmart.html
0002 Walmart.html
0007 eBay.html
0008 Walmart.html
0010 eBay.html
0011 Walmart.html
0015 amz.html
0016 Walmart.html
0017 Walmart.html
0018 Walmart.html
```

It'd be helpful if there were a command-line tool that output the list of files in a particular directory as a multiline string, so that the list could be copied and pasted for input into the Vectorizer. For example something like `fathom-pick list training` might output such a multiline string for all the files in the `training` directory.

Currently, I have been using `ls > ~/output.txt`, opening this file in a text editor and copying and pasting it from there.

erikrose:
On the Mac, you can just select the files in the Finder, Copy, and Paste into BBEdit to get the list of names. However, I see that other text editors act differently. vi, for example, lists the entire paths. And Daniel wasn't able to find similar functionality readily with Windows Explorer + Notepad. So this could be a nice addition (`fathom-list`?) if https://github.com/mozilla/fathom-fox/issues/31 can't do it more straightforwardly. (In fact, such a util came up during one of my spin-up sessions with Daniel.) Thanks!

danielhertenstein:
Another option is changing how the vecorizer works such that you don't need to input filenames. We could look at the directory specified in the `Base URL:` text box and vectorize all HTML files we see. We would be making an assumption about the user's file structure, but we already expect a certain file structure from using `fathom-pick`.

erikrose:
A couple cautions come to mind:
* Sometimes you want to vectorize only a few files, just to test a hypothesis, especially when experimenting on large corpora.
* By crawling the Python-served index page (which I assume is what you mean), we'd be coding in the assumption that the index is served in the format that Python's `http.server` happens to use. I have a feeling these corpora may someday be on a remote server and served by something else.
* fathom-pick doesn't mandate or even suggest a file structure; it just moves a random selection of files. Perhaps you're thinking of the recommended FS layout from the training docs.

danielhertenstein:
Those are all good cautions. It would be simple enough to make a `fathom-list` tool.

danielhertenstein:
[...]

PR #100, opened 2019-07-01
Adds commandline tool to list filenames for vectorization and optionally save to a file

danielhertenstein:
Addresses #97 

This PR adds a command line tool to aid in vectorization. You can list the files in a directory for easy copy-pasting into the vectorizer box in Fathom-Fox. Optionally, you can save the output to a file so you can reference the file later and forgo using this tool every time.

### DIFF
--- a/cli/fathom_web/list.py
+++ b/cli/fathom_web/list.py
@@ -1,18 +1,19 @@
-from click import argument, command, File, option, Path
 import pathlib
+
+from click import argument, command, File, option, Path
 
 
 @command()
-@option('--out_file', '-o', type=File(mode='w'), default=None,
+@option('--out-file', '-o', type=File(mode='w'), default=None,
         help='A file for saving the printed filenames for easy future reference.')
-@argument('in_directory', type=Path(exists=True, file_okay=False, allow_dash=True))
+@argument('in_directory', type=Path(exists=True, file_okay=False))
 def main(in_directory, out_file):
     """
     Lists filenames in a IN_DIRECTORY, one filename per line. Optionally saves output to OUT_FILE.
 
     Ignores hidden files.
 
-    This is useful for vectorizing samples using fathom-fox. Fathom-fox expects input filenames copied into a text box
+    This is useful for vectorizing samples using FathomFox. FathomFox expects input filenames copied into a text box
     with one filename per line.
     """
     if out_file is not None:
--- a/cli/fathom_web/list.py
+++ b/cli/fathom_web/list.py
@@ -1,0 +1,32 @@
+from click import argument, command, File, option, Path
+import pathlib
+
+
+@command()
+@option('--out_file', '-o', type=File(mode='w'), default=None,
+        help='A file for saving the printed filenames for easy future reference.')
+@argument('in_directory', type=Path(exists=True, file_okay=False, allow_dash=True))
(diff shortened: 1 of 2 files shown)


cli/fathom_web/list.py
```diff
@@ -0,0 +1,32 @@
+from click import argument, command, File, option, Path
+import pathlib
+
+
+@command()
+@option('--out_file', '-o', type=File(mode='w'), default=None,
```
erikrose: If we change this to `--out-file` (with a hyphen), we'll be consistent with fathom-train and the other tools. `click` will change it to an underscore when stuffing its value in a variable.

cli/fathom_web/list.py
```diff
@@ -0,0 +1,32 @@
...
+        help='A file for saving the printed filenames for easy future reference.')
+@argument('in_directory', type=Path(exists=True, file_okay=False, allow_dash=True))
+def main(in_directory, out_file):
+    """
+    Lists filenames in a IN_DIRECTORY, one filename per line. Optionally saves output to OUT_FILE.
+
+    Ignores hidden files.
+
+    This is useful for vectorizing samples using fathom-fox. Fathom-fox expects input filenames copied into a text box
```
erikrose: FathomFox

cli/fathom_web/list.py
```diff
@@ -0,0 +1,32 @@
...
+        if file.name.startswith('.'):
+            continue
+
+        print(file.name)
+
+        if out_file is not None:
+            filenames_to_save.append(file.name + '\n')
+
+    if out_file is not None:
+        out_file.writelines(filenames_to_save)
```
erikrose: Where ever did you find this `writelines` routine? click doesn't document it, and the only one I can find in Python 3 is in asyncio. Its docs say we should be calling drain() afterward. Should this be a higher-level routine, or should I be enlightened?

cli/fathom_web/list.py
```diff
@@ -0,0 +1,32 @@
+from click import argument, command, File, option, Path
+import pathlib
+
+
+@command()
+@option('--out_file', '-o', type=File(mode='w'), default=None,
+        help='A file for saving the printed filenames for easy future reference.')
+@argument('in_directory', type=Path(exists=True, file_okay=False, allow_dash=True))
```
erikrose: I'm not sure that allow_dash is applicable for a dir.

cli/fathom_web/list.py
erikrose: I am enlightened!

cli/fathom_web/list.py
```diff
@@ -0,0 +1,32 @@
+from click import argument, command, File, option, Path
+import pathlib
```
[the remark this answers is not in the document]
danielhertenstein: Will do!

cli/fathom_web/list.py
danielhertenstein: Will do!

cli/fathom_web/list.py
danielhertenstein: Whoops. I went by the repo name instead of checking the official name.

cli/fathom_web/list.py
danielhertenstein: As we spoke about, I will remove it.

Issue #104, opened 2019-07-03
Trainer was slow for Price Tracker


danielhertenstein:
When @biancadanforth and I were training the coefficients for Price Tracker, ~2500 iterations took ~25 minutes. This was surprising to @erikrose. We should investigate this.

danielhertenstein:
Using the Price Tracker image coefficients case (which early stops around 2500 iterations) as a test case, I obtained an overall profile for the run.
![overall_profile](https://user-images.githubusercontent.com/5586735/60617754-64a04a80-9da2-11e9-9672-a295a1cbda3c.PNG)

It seemed odd that computing the accuracy per tag was 99% of the runtime, so I did a 1000 iteration run of the same vectors to obtain a line by line profile.
![accuracy_per_tag_profile](https://user-images.githubusercontent.com/5586735/60617849-9b766080-9da2-11e9-888c-003e5528b074.PNG)

You can see on line 90, there is a loop running each sample through the model one at a time. We should instead evaluate all the samples at once as is done in the training loop.

I have a pull request coming to address this.

danielhertenstein:
Closed with #105

PR #105, opened 2019-07-03
Accuracy per tag optimization

danielhertenstein:
Addresses #104 

Previously, `accuracy_per_tag()` ran each sample through the model one at a time and then checked the accuracy for that sample. This pull request runs all the samples through the model at the same time and then computes the accuracy metrics for all the samples using numpy methods.

There is an additional change to `accuracy_per_tag()` such that it now takes predictions as an argument instead of the model and the samples. This is because the main source of calls for `accuracy_per_tag()` is `learn()` which itself runs all the samples through the model and therefore has the predictions laying around for use. This change requires the last calls to `accuracy_per_tag()`, located within `main()`, to run the samples through the model beforehand.

The new overall profile shows that the overall runtime and the runtime for `accuracy_per_tag()` are much lower.
new_overall_profile

### DIFF
--- a/cli/fathom_web/train.py
+++ b/cli/fathom_web/train.py
@@ -294,12 +294,10 @@ def main(training_file, validation_file, stop_early, learning_rate, iterations, 
         validation_arg = None
     model = learn(learning_rate, iterations, x, y, validation=validation_arg, stop_early=stop_early, run_comment=full_comment)
     print(pretty_coeffs(model, training_data['header']['featureNames']))
-    y_pred = model(x)  # Make predictions.
-    accuracy, false_positive, false_negative = accuracy_per_tag(y, y_pred)
+    accuracy, false_positive, false_negative = accuracy_per_tag(y, model(x))
     print(pretty_accuracy(('  ' if validation_file else '') + 'Training accuracy per tag: ', accuracy, len(x), false_positive, false_negative))
     if validation_file:
-        validation_preds = model(validation_ins)
-        accuracy, false_positive, false_negative = accuracy_per_tag(validation_outs, validation_preds)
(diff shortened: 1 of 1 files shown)


cli/fathom_web/train.py
```diff
@@ -290,10 +294,12 @@ def main(training_file, validation_file, stop_early, learning_rate, iterations,
         validation_arg = None
     model = learn(learning_rate, iterations, x, y, validation=validation_arg, stop_early=stop_early, run_comment=full_comment)
     print(pretty_coeffs(model, training_data['header']['featureNames']))
-    accuracy, false_positive, false_negative = accuracy_per_tag(model, x, y)
+    y_pred = model(x)  # Make predictions.
```
erikrose: I actually slightly prefer inlining `model(x)` into the following function call (and the same below, for validation). I don't think it's any harder to comprehend (a look at the name of the param it's getting passed into in accuracy_per_tag() clues you in if not), and it doesn't add more state to an already pretty stateful procedure. What do you think?

cli/fathom_web/train.py
```diff
@@ -82,18 +83,21 @@ def learn(learning_rate, iterations, x, y, validation=None, stop_early=False, ru
     return model
 
 
-def accuracy_per_tag(model, x, y):
+def accuracy_per_tag(y, y_pred):
     """Return the accuracy 0..1 of the model on a per-tag basis, given input
```
erikrose: Docstring needs an update since the params changed.

cli/fathom_web/train.py
danielhertenstein: I'm on board with this.

cli/fathom_web/train.py
danielhertenstein: Whoops. I will update that.

Issue #102, opened 2019-07-01
Store page subresources in separate files


erikrose:
GitHub gets grumpy about us uploading >2GB git packs (i.e. it refuses), making it hard to share ruleset repos. Let's split page subresources (images, etc.) out and store them in git-lfs.

Pros:
+ Simple simple simple to use.
+ Versioned with code.
+ History-preserving
+ No 200MB file-size cap
+ Per-project permissions. No new ACLs to set up.

Cons:
- No diffs.

But, if we store the subresources in LFS and commit the markup to plain git, then we can diff it after all (and much faster than now), and the cons go to nil. git-lfs-migrate can even help automate this. LFS costs a little money but not enough to bother about at GitHub.

2 spot-checks measure 123x (on a 9.8MB sample) and 877x (on a 98MB sample) reductions in size by stripping images and other subresources out. (Files stored in LFS don't count toward your quota at GitHub.) That means we'll have room for about 10,000 samples, which should be plenty for the foreseeable future. If awesome RNN advances mean more than that is worthwhile to collect, I'll be thrilled to come up with a fancier solution at that point.

We'll need **a tool to extract subresources from existing samples** and **a plan for making them external on future ones**. I currently suspect we'll do the same freeze-and-extract dance on future ones, since otherwise we'll need some way of downloading multiple files from freeze-dry, like tarring them up, while still keeping all the files that go with one page together, in case we're using the bulk Corpus Collector.

danielhertenstein:
I will start with the login-forms corpus.

danielhertenstein:
@erikrose, what are we considering subresources? Anything that is a link out on the page normally?

erikrose:
Anything currently inlined as a data URL: images, stylesheets, iframes, and so on

danielhertenstein:
Do we care about converting the base64 strings back to their original formats after extracting them? If we're just going to compress them and stick them in LFS, I don't see a need to convert them.

erikrose:
De-base-64-ing them is the only way I can think of to get the pages to load them again: just stick them in the (LFS'd) repo as images and tweak the HTML to reference them as normal files. Can you think of another way?

danielhertenstein:
The other option is to require a person to run another script that will reinsert the base64 strings before they can load the page again. But it is simple enough to convert them so they can be referenced and I will go that route. I wanted to make sure we had a need for it before I continued down that path.

erikrose:
[...]

PR #109, opened 2019-07-12
Adds a tool to extract resources from freeze-dry'd html files

danielhertenstein:
This PR adds a tool called `fathom-extract` which will extract all base64 strings out of a page saved by freezeDry, save them in a new directory, and point the html to those new files.

This tool will allow us to save our corpora in git (and git LFS).

There is an option, on by default, to preserve the original files which I added because I'm still a little scared =P


cli/fathom_web/extract.py
```diff
@@ -0,0 +1,128 @@
+import base64
+import mimetypes
+import pathlib
+import re
+
+from click import argument, command, Path
+
+
+BASE64_DATA_PATTERN = re.compile(r'(data:(?P<mime>[a-zA-Z0-9]+?/[a-zA-Z0-9\-.+]+?);(\s?charset=utf-8;)?base64,(?P<string>[a-zA-Z0-9+/=]+))')
```
erikrose: Maybe one of your problem pages actually has a "data:blah;blah:base64"-looking string in it that isn't actually a data attr and it's throwing things off? Just a casual idea.

erikrose: I wonder, what made you abandon Beautiful Soup for a grep-based approach? I worry a little about a reduction in precision.

danielhertenstein: I abandoned Beautiful Soup because the page visually changed when doing nothing to the data regardless of which parser I used during input and how I saved the data during output. For example, the code:
```
# image original.png
with open ('267.html') as fp:
    soup = BeautifulSoup(fp)

# image str.png    
with open('new_str.html', 'w') as fp:
    fp.write(str(soup))

# image prettified.png
with open('new_pret.html', 'w') as fp:
    fp.write(soup.prettify())
```
results in the following three pages (in the same order as the code):
original
In str.png, �'s appear next to the phone number at the top and on the home button
str
In prettified.png, the home button gets messed up and all of the `|`'s shift
prettified

danielhertenstein: I do think there could be problems with my regex, though it was necessary in the Beautiful Soup version anyway because some of the base64 strings are contained in string attributes of tags that contain a lot of other text. For example in 267.html, one of the base64 strings is here:

```
        <style>
        .top-link-cart{
			background: url(data:image/png;base64,iVBORw...ErkJggg==)  no-repeat scroll left center transparent !important;
			
			 padding:13px 6px 13px 40px !important;
		}
        
        </style>
```
and Beautiful Soup throws everything inside this style tag into a `Tag.string` attribute.

danielhertenstein: Okay, I have found the two changes that solve the remaining problems.

1. Pages with a `<base>` tag need that tag removed.
2. The other two pages (and three of the pages with a `<base>` tag) require setting privacy.file_unique_origin to false in about:config

cors
https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSRequestNotHttp

For the `<base>` tag, it's still not clear to me why the tag is added.
For the CORS problem, I hope there is a better solution, but I haven't found one yet.

erikrose: Aha. Thanks! I'd be mildly curious to diff the 2 outputted files and see what the changes are.

danielhertenstein:
For reference, so it can be linked back to the issue, this addresses #102

cli/setup.py
```diff
@@ -14,11 +14,12 @@
     url='https://mozilla.github.io/fathom/',
     install_requires=['click>=7.0,<8.0', 'tensorboardX>=1.6,<2.0', 'torch>=1.0,<2.0'],
     entry_points={'console_scripts': [
+        'fathom-extract = fathom_web.extract:main'
```
danielhertenstein: I think we should make these alphabetical

danielhertenstein:
Okay, as far as I can tell, this does everything it needs to do now. Note that you may want to use #117 to serve the pages so you can see that they load correctly.

cli/setup.py
erikrose: Yes please!

erikrose: Also, the missing comma at the end of the line here wreaks havoc. :-)

cli/fathom_web/extract.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+@command()
+@option('--preserve-originals/--no-preserve-originals',
+        default=True,
+        help='Save original html files in a newly created `originals` directory in IN_DIRECTORY (default: True)')
+@argument('in_directory', type=Path(exists=True, file_okay=False))
+def main(in_directory, preserve_originals):
+    """
+    Extracts resources from the html pages in IN_DIRECTORY and stores them in a separate directory for Git-LFS storage.
```
erikrose: Where is that second directory? I'd want to know before running it if I was a first-time user (which I am).

cli/fathom_web/extract.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+    if preserve_originals:
+        originals_dir = pathlib.Path(in_directory) / 'originals'
+        originals_dir.mkdir(parents=True, exist_ok=True)
+
+    for file in pathlib.Path(in_directory).iterdir():
+        if file.suffix != '.html':
+            print(f'Skipping {file.name}; not an html file')
```
erikrose: I get "Skipping originals; not an html file", which is kind of silly. :-)

cli/fathom_web/extract.py
```diff
@@ -0,0 +1,144 @@
+import base64
+import mimetypes
+import pathlib
+import shutil
+import re
+
+from click import argument, command, option, Path
+
+
+BASE64_DATA_PATTERN = re.compile(r'(data:(?P<mime>[a-zA-Z0-9]+?/[a-zA-Z0-9\-.+]+?);(\s?charset=utf-8;)?base64,(?P<string>[a-zA-Z0-9+/=]+))')
```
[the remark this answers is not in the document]
erikrose: Just tested it. Boy, it's really fast even as you have it!

cli/fathom_web/extract.py
```diff
@@ -0,0 +1,144 @@
...
+    We do this by building a new html string using the non-base64 data pieces from the original file and the
+    filenames we will generate for each of the base64 data strings.
+
+    Base64 data is found with regex matching. Each data string is decoded and saved as a separate file.
+    """
+    with file.open(encoding='utf-8') as fp:
+        html = fp.read()
+
+    # Make the subresources directory
+    subresources_directory = file.parent / 'resources' / f'{file.stem}_resources'
```
erikrose: Is there a purpose for the "_resources" suffix? If not, I'd do without to save a little reading, typing, and space. It's already within a folder called "resources".

cli/fathom_web/extract.py
```diff
@@ -0,0 +1,144 @@
...
+    if preserve_originals:
+        originals_dir = pathlib.Path(in_directory) / 'originals'
+        originals_dir.mkdir(parents=True, exist_ok=True)
+
+    for file in pathlib.Path(in_directory).iterdir():
+        if file.suffix != '.html':
+            print(f'Skipping {file.name}; not an html file')
+            continue
+        if preserve_originals:
+            shutil.copyfile(file, originals_dir / file.name)
```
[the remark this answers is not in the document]
erikrose: Also, it'd save IO to *move* the original and then re-write the new (smaller) thing in the old place.

cli/fathom_web/extract.py
```diff
@@ -0,0 +1,144 @@
...
+
+    base64_data_matches = BASE64_DATA_PATTERN.finditer(html)
+    for match in base64_data_matches:
+        # Add the content before the base64 data
+        new_html += html[offset:match.start(1)]
+
+        base64_string = match.group('string')
+
+        # Check to see if we have already encountered this base64 string
+        file_path = saved_strings.get(base64_string, None)
```
erikrose: `None` is the default, so we don't have to specify it.

cli/fathom_web/extract.py
```diff
@@ -0,0 +1,144 @@
...
+            mime_type = match.group('mime')
+            filename_counter += 1
+            filename = generate_filename(mime_type, str(filename_counter))
+            binary_data = decode_base64_string_to_binary(base64_string)
+            file_path = subresources_directory / filename
+            save_binary_data(binary_data, file_path)
+            saved_strings[base64_string] = file_path
+
+        # "Replace" the old base64 data with the relative path to the newly created file
+        new_html += file_path.relative_to(file.parent).as_posix()
```
erikrose: I suppose we'll have to urlencode this, too, in case anybody has, say, a space in their FS path.

cli/fathom_web/extract.py
erikrose: The drag is that, if we stick the `resources` dir next to the samples being processed, we can't really move the samples afterward without breaking the references. I envision us running fathom-extract as a first step, as soon as a corpus is collected, because otherwise we can't check it in anywhere. Then, when we use fathom-pick to grab out some training samples, their resources won't come with them. The obvious solutions I see are…

1. Put a single resources dir somewhere, like at the root of the samples dir. Of course, this means we can never move samples any shallower or deeper in the folder hierarchy.
2. Do (1), but also stick a symlink to the root-level resources dir in each sample dir (training, validation, etc.). Git will store symlinks cluefully, but I'm not sure how they check out on Windows.
3. Do (1), but teach fathom-serve about a special `resources` dir (still living at the root level), and have it expose said dir everywhere, so validation/resources will work, and training/resources will work, but they'll point to the same dir on disk. This will complicate things for the user, because they'll either need to pass fathom-serve the location of the resources dir, or fathom-serve will have to track it down itself according to some conventions, which will surprise someone at some point.

My favorite is (2), at least if it works on Windows. I've used a similar trick for letting PHP apps find their shared includes, and it worked well.

Any other alternatives?

cli/fathom_web/extract.py
```diff
@@ -0,0 +1,144 @@
...
+    # `mimetypes` gets some extensions wrong (e.g. image/jpeg -> .jpe) and doesn't
+    # work for some MIME types that freeze-dry gives so use our own mapping first
+    try:
+        extension = MIME_TYPE_TO_FILE_EXTENSION[mime_type]
+    except KeyError:
+        extension = mimetypes.guess_extension(mime_type, strict=True)
+    return f'{filename}{extension}'
+
+
+def save_binary_data(binary_data: bytes, filepath: pathlib.Path):
```
erikrose: Do this and decode_base64_string_to_binary exist as separate functions just to give you a place to hang docstrings?

cli/fathom_web/extract.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+    return base64.b64decode(base64_string)
+
+
+def generate_filename(mime_type: str, filename: str) -> str:
+    """
+    Creates a filename to use for saving the base64 data with the appropriate file extension.
+
+    We can't necessarily get an appropriate filename from the html the base64 data comes from, so we don't
+    even try. Instead we just an incrementing counter to ensure unique filenames.
```
erikrose: "we just an"

cli/fathom_web/extract.py
erikrose: Let's be consistent w.r.t. verb forms. In extract_base64_data_from_html_page(), you say "Extract", but here it's "Extracts". I favor the former, based on https://www.python.org/dev/peps/pep-0257/#one-line-docstrings, which is what I've done elsewhere in the codebase.

cli/fathom_web/extract.py
```diff
@@ -0,0 +1,144 @@
...
+        if preserve_originals:
+            shutil.copyfile(file, originals_dir / file.name)
+        extract_base64_data_from_html_page(file)
+
+
+def extract_base64_data_from_html_page(file: pathlib.Path):
+    """
+    Extract all base64 data from the given html page and store the data in separate files.
+
+    We do this by building a new html string using the non-base64 data pieces from the original file and the
```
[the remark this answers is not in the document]
erikrose: html→HTML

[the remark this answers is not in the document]
danielhertenstein: Will do!

cli/fathom_web/extract.py
danielhertenstein: Wow, forgot about that. Thanks.

cli/fathom_web/extract.py
danielhertenstein: The "Extract" form is cool with me. Seems like the "Extracts" form is what I default to, so I'll work on retraining myself. Apologies in advance when I miss one or three in the future.

cli/fathom_web/extract.py
danielhertenstein: Thanks!

cli/fathom_web/extract.py
danielhertenstein: I'll put a separate check and message for directories.

cli/fathom_web/extract.py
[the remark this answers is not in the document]
danielhertenstein: When you say:
> It would be spiffy if this didn't overwriting existing files.

do you mean you'd prefer a new file always be made or are you saying (along with the second comment here) that you'd rather this code path move the file instead of copy?

cli/fathom_web/extract.py
danielhertenstein: That is a good point. I can update this when we decide what to do regarding your comment below on the various linking options.

cli/fathom_web/extract.py
danielhertenstein: `decode_base64_string_to_binary()` changed throughout developing this and wasn't originally a one-liner. I can inline that. As for `save_binary_data()`, I tend to pull out pieces like this that add indentation levels so the main code (in this instance the body of the `for` loop starting on line 80) reads easily at a high level. But, I do have the saving of the file on lines 106 and 107 right there in the code, so I suppose I should be fine inlining this as well.

erikrose:
Yes, I had to change my default, too. :-)

cli/fathom_web/extract.py
erikrose: The weird part was just that fathom-extract had itself created that dir, so it was strange that it then turned around and complained about it.

cli/fathom_web/extract.py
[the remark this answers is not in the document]
erikrose: Neither, actually! What I meant was, since, when you create the originals dir, you say it's okay if it already exists, it makes me nervous to go around overwriting the contents, since it could be that somebody manually made an originals dir and was sticking conflictingly-named things in it. What I would do is crash if there is an existing file there with a conflicting name. Or, if you're feeling fancy, start sticking numbers after the filename until it no longer conflicts (`313 2.html`, `313 3.html`, etc.). But it's not clear to me that the latter is even preferable, so I'd just let it crash, ideally with a nice message.

cli/setup.py
danielhertenstein: Darn, I don't know how I missed this. Thanks for catching it!

cli/fathom_web/extract.py
erikrose: I don't have a strong opinion. I've written the same function before. In my case, I did it because I was writing a whole bunch of literals to various files, and it saved a line for each one.

erikrose:
Couldn't figure out for 5 minutes why it wasn't working! :-)

cli/fathom_web/extract.py
erikrose: Hmm, symlinks on Windows do not look promising: https://github.com/git-for-windows/git/wiki/Symbolic-Links. (I know we experimented with it a few weeks back, but I forgot how it worked out.) But wait, maybe they just started working! https://www.quora.com/How-does-Git-handle-symbolic-links-on-Windows

cli/fathom_web/extract.py
danielhertenstein: Upon inlining, it looks clear enough to me, so that's what will happen in the next commit.

cli/fathom_web/extract.py
danielhertenstein: Yup, you are right.

cli/fathom_web/extract.py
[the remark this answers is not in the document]
danielhertenstein: I get you now. I also like saying it's not okay if the `originals` directory already exists. It feels the safest.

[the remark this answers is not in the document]
erikrose: Great. Easiest, too!

cli/fathom_web/extract.py
danielhertenstein: The quora page you linked makes it seem there are still significant hoops a Windows user would have to jump through for symlinks to work. What do you think about storing resources in separate directories next to the pages? We could change `fathom-pick` to move resource directories. It will probably make setting up a repo and git-LFS a little more difficult, but I imagine we'll already need to supply a guide for users for that.

erikrose: We decided to make changes to fathom-pick to also move resources. This means Windows people don't need to mess with their machines and we still can just double-click samples to open them if you futz with your FF CORS settings.

cli/fathom_web/extract.py
```diff
@@ -58,11 +57,17 @@ def main(in_directory, preserve_originals):
         if file.suffix != '.html':
             print(f'Skipping {file.name}; not an HTML file')
             continue
-        extract_base64_data_from_html_page(file, preserve_originals, originals_dir)
 
+        html = extract_base64_data_from_html_page(file)
```
erikrose: Much better factoring. Yay for fewer, shallower side effects! :-)

cli/fathom_web/extract.py
```diff
@@ -83,7 +83,7 @@ def extract_base64_data_from_html_page(file: pathlib.Path):
         html = fp.read()
 
     # Make the subresources directory
-    subresources_directory = file.parent / 'resources' / f'{file.stem}_resources'
+    subresources_directory = file.parent / f'{file.stem}_resources'
```
erikrose: Oh, I think we had a miscommunication. I thought the outcome of yesterday's conversation was that we'd keep the `samples/negatives/resources/45/2.png`-like layout and just remove the "_resources" suffix.

danielhertenstein: Indeed we did. I think this stems out of how we think `fathom-pick` will work now. Will `fathom-pick` create a `resources` directory in your destination directory and move the page-specific resource directories for the randomly picked items into that new `resources` directory?

erikrose: Yep. And if the dir is already there, it'll just add stuff to it. But if the sample-specific subdir is already there, it should error.

danielhertenstein: Gotcha. I'm on it!

erikrose:
Thanks, Daniel! :-D

cli/fathom_web/extract.py
```diff
@@ -36,10 +36,11 @@
 def main(in_directory, preserve_originals):
     """
     Extract resources from the HTML pages in IN_DIRECTORY and store the
-    resources for each page in a newly created directory in IN_DIRECTORY.
+    resources for each page in a newly created page-specific directory
+    within a newly created resources directory in IN_DIRECTORY.
     For example, the resources for `example.html` would be stored in
-    `example_resources/`. This tool is used to prepare your samples for a
-    git-LFS enabled repository.
+    `resources/example_resources/`. This tool is used to prepare your
```
erikrose: Turns out you need 2 backticks on either side in ReStructured Text!

danielhertenstein: Ah yes. Thank you for your great attention to detail!

PR #110, opened 2019-07-16
Documentation updates

Baron-Severin:
A few notes that came up during a meeting with @erikrose.

If these updates are not entirely accurate, or formatted in a way that you would not prefer, I will be happy to make any requested changes.

### DIFF
--- a/docs/clustering.rst
+++ b/docs/clustering.rst
@@ -2,7 +2,9 @@
 Clustering
 ==========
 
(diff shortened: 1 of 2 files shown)


docs/clustering.rst
```diff
@@ -2,7 +2,9 @@
 Clustering
 ==========
 
-One of Fathom's most powerful features is its flexible clustering algorithm, which groups nodes based on their proximity and ancestry. It is documented here as top-level functions but is also available directly within rulesets as :func:`bestCluster`, which has the advantage of letting you direct its results to further rules. Clustering is computationally expensive (at least O(n^2)) but widely applicable.
```
Baron-Severin: This splits off the final sentence ("Clustering is computationally expensive (at least O(n^2)) but widely applicable.") into a separate note, and expands upon it slightly.  Otherwise, the line is untouched.

erikrose:
Thanks, Severin! :-D

PR #111, opened 2019-07-16
Improve docs on page() function.

erikrose:
Also replace `node` with `fnode`, which accurately represents the type of the arg.


erikrose:
That's actually intentional! "[Iff"](https://en.wikipedia.org/wiki/If_and_only_if) is an abbreviation for "if any only if".

PR #112, opened 2019-07-16
Add fathom-test tool. Fix #90.

### LINKED ISSUE #90: Add testing-accuracy measurement to trainer
At the moment, `fathom-train` reports training and validation accuracy but not testing accuracy. Design a way to feed a trained set of coeffs and biases into the trainer and measure the accuracy that results.

erikrose:
* Move Python commands down a level into a `commands` folder.
* Factor out common bits I need from fathom-train.

It turned out to have so many differences from the trainer that it warranted its own tool. Basically none of the commandline interface is the same, all the fathom-train options having to do with training itself. What I did instead is to factor out the common stuff: mostly accuracy calculation and pretty-printing, into an `accuracy` module. The new code is all in test.py; that's what needs review. Also, I'd like feedback on the user-facing design of the tool. One known issue is that `click` makes it hard to pretty-print the JSON in the --help message. I didn't think it worth it to make that perfect now, as it might involve subclassing a bunch of click itself.

To get a quick runthough, you can grab the coeffs out of the login-forms trainees.js and use any old training or validation vector file as a set of pretend testing vectors:

```
% fathom-test /Users/erose/Checkouts/login-forms/vectors/trainingUsernameRegisterAttrsOrContent.json  '{"coeffs": [
        ["emailKeywords", 0.3606211543083191],
        ["loginKeywords", 5.311713218688965],
        ["headerRegistrationKeywords", -1.6875461339950562],
        ["buttonRegistrationKeywordsGte1", -2.22440767288208],
        ["formPasswordFieldsGte2", -6.207341194152832],
        ["formTextFields", -1.3702701330184937]
    ],
 "bias": -34.7013704776763916}'

Testing accuracy per tag:  0.78307    95% CI: (0.72431, 0.84183)  FP: 0.000  FN: 0.217
Testing accuracy per page: 1.00000    95% CI: (1.00000, 1.00000)

Testing per-page results:
 success  on   334.html. Confidence: 0.00000000
 success  on     1.html. Confidence: 0.00000000
 success  on     5.html. Confidence: 0.00000000
 success  on    14.html. Confidence: 0.96090430
…
```

I've tweaked fathom-train to output the coeffs and biases in a form that can be just copied and pasted into fathom-test.


cli/fathom_web/commands/test.py
```diff
@@ -0,0 +1,77 @@
...
+
+def decode_weights(ctx, param, value):
+    """Validate a click option, making sure it's a valid JSON object with
+    properly formatted "coeff" and "bias" keys."""
+    try:
+        decoded_weights = loads(value)
+    except JSONDecodeError:
+        raise BadParameter('Weights must be a valid JSON object.')
+    else:
+        if 'coeffs' in decoded_weights and 'bias' in decoded_weights:
```
danielhertenstein: You can avoid this amount of nesting by checking for the bad situations and raising exceptions.
```
if 'coeffs' not in decoded_weights and 'bias' not in decoded_weights:
    raise BadParameter('Weights must contain "coeffs" and "bias" keys.')
if not isinstance(decoded_weights['bias'], float):
    raise BadParameter('Bias must be a float.')
if not isinstance(decoded_weights['coeffs'], list) or any((len(pair) != 2 or not isinstance(pair[0], str) or not isinstance(pair[1], float)) for pair in decoded_weights['coeffs']):
    raise BadParameter('Coeffs must be a list of 2-element lists: [["ruleName", numericCoefficient], ...].')
return decoded_weights
```

cli/fathom_web/commands/test.py
```diff
@@ -0,0 +1,77 @@
...
+            if type(decoded_weights['bias']) is float:
```
danielhertenstein: I believe the best practice for type checking is through `isinstance()` ([this](https://stackoverflow.com/questions/1549801/what-are-the-differences-between-type-and-isinstance) is my goto post on the topic) instead of `type` unless you want to be checking for exact types.

cli/fathom_web/commands/test.py
```diff
@@ -0,0 +1,77 @@
...
+                if (type(decoded_weights['coeffs']) is list and
+                    all((len(pair) == 2 and type(pair[0]) is str and type(pair[1]) is float)
```
danielhertenstein: This if-clause is a bit unwieldy. Pulling the whole thing out into its own method would make it easier to read at the expense of the additional lines. I think this is personal preference.
```
def coeffs_not_list_of_2_element_lists(coeffs):
    if not isinstance(coeffs, list):
        return False
    for pair in coeffs:
        if len(pair) != 2:
            return False
        if not isinstance(pair[0], str):
            return False
        if not isinstance(pair[1], float):
            return False
    return True

...

if coeffs_not_list_of_2_element_lists(decoded_weights['coeffs']):
    raise BadParameter('Coeffs must be a list of 2-element lists: [["ruleName", numericCoefficient], ...].')
```

cli/fathom_web/commands/test.py
erikrose: You are correct! Sorry, my JS is seeping back into my Python. Eek!

cli/fathom_web/commands/test.py
erikrose: There is a bit of personal preference here, I think. I tend to not mind nesting or width in general and instead dislike early exits, because they make me read not only the whitespace but also scan for about 5 different keywords (break, continue, return, raise, etc.). But I'm fine flattening this out if you prefer, as the structure is nice and regular. I mostly used your code, but I flopped the final nots with De Morgan's law to avoid having quite so many negations, as those exceed my brain's capacity. :-) Patch to follow.

cli/fathom_web/commands/test.py
erikrose: I rewrapped the predicates in the generator expression. Hopefully that makes it clearer?

erikrose:
Thanks for the good review! How's that?

erikrose:
The Python kit is just teetering on the edge of needing tests. There are a lot of branches in decode_weights (which, honestly, I wrote to that degree of thoroughness recreationally), and I grow tired of running through them by hand. And the team is getting bigger, so having to actually run other people's stuff by hand will become a time sink. It's only the late hour and the thought of wiring another language's test fixtures into the Travis config and debugging it that kept me from doing it. Next time, though.

danielhertenstein:
All the changes look good to me! I agree that we should have tests. I'm happy to divide and conquer covering the tools.

PR #115, opened 2019-07-19
Update packages

biancadanforth:


biancadanforth:
>  But shouldn't there also be updates to package.json?

@erikrose : Actually, I am not surprised that there are only updates to `package-lock.json`, after talking with Mythmon (TIL, heh).

Many if not most security problems are nested dependencies, and those aren't specified in `package.json`. They are pinned in `package-lock.json`, so that’s the place they need to be updated.

Most libraries don't specify exact versions for their dependencies -- instead they specify a [semver range](https://docs.npmjs.com/misc/semver#versions).

Taking for example the `js-yaml` nested dependency that is updated from v3.12.0 to v3.13.1 in this PR:
- With `npm list js-yaml`, I see that it's a nested dependency, among other packages, for `coveralls`
- We currently use `coveralls` v3.0.2.
- According to [`coveralls`’ v3.0.2 `package.json` for `js-yaml`](https://github.com/nickmerwin/node-coveralls/blob/3.0.2/package.json#L34), `coveralls` v3.0.2 lists `"js-yaml": "^3.11.0",` under `dependencies`.
- This means that the v3.0.2 version of `coveralls` is compatible with any version of `js-yaml` >= v3.11.0 and < v4, which includes v3.13.1.

erikrose:
Oh yes, I was aware of all that. I just figured if you were putting up an "upgrade packages" PR, you might as well up the versions of our direct dependencies too. :-)

biancadanforth:
I'm happy to include general updates if you like, though that wasn't the intention here. I will add another commit.

erikrose:
Tests pass, so lgtm! Thanks, Bianca!

PR #116, opened 2019-07-19
Fix #118: Improve `isVisible` for correctness and performance

### LINKED ISSUE #118: Improve `isVisible` for correctness and performance
Related to #91 .

As mentioned in this [Price Tracker issue](https://github.com/mozilla/price-tracker/issues/319), `isVisible` overwhelmingly contributes to excessive jank right around page load.

Since Fathom's `isVisible` method is largely based on Price Tracker's, it suffers from this same problem, and it would be a good idea to have a more performant implementation.

Additionally, the current implementation in Fathom [has a bug](https://github.com/mozilla/price-tracker/issues/319#issuecomment-513313847), so this should be fixed as well.

biancadanforth:
The width and height properties in the return value from getComputedStyle are of the form '*px'.

### DIFF
--- a/utilsForFrontend.mjs
+++ b/utilsForFrontend.mjs
@@ -458,30 +458,47 @@ export function sigmoid(x) {
 }
 
 /**
- * Return whether an element is practically visible, considing things like 0
- * size or opacity, and clickability.
+ * Return whether an element is practically visible, considering things like 0
+ * size or opacity, ``visibility: hidden`` and ``overflow: hidden``.
+ *
+ * This could be 5x more efficient if https://github.com/w3c/csswg-drafts/issues/4122
+ * happens.
  */
 export function isVisible(fnodeOrElement) {
     const element = toDomElement(fnodeOrElement);
-    const rect = element.getBoundingClientRect();
-    if (rect.width === 0 || rect.height === 0) {
+    // Avoid reading ``display: none`` due to Bug 1381071
+    const elementRect = element.getBoundingClientRect();
+    if (elementRect.width === 0 && elementRect.height === 0) {
         return false;
     }
-    const style = getComputedStyle(element);
-    if (style.opacity === '0') {
+    const elementStyle = getComputedStyle(element);
+    if (elementStyle.visibility === 'hidden') {
         return false;
     }
-    // workaround for https://github.com/w3c/csswg-drafts/issues/4122
-    const scrollX = window.pageXOffset;
-    const scrollY = window.pageYOffset;
-    const absX = rect.x + scrollX;
-    const absY = rect.y + scrollY;
-    window.scrollTo(absX, absY);
-    const newX = absX - window.pageXOffset;
-    const newY = absY - window.pageYOffset;
-    const eles = document.elementsFromPoint(newX, newY);
-    window.scrollTo(scrollX, scrollY);
-    return eles.includes(element);
+    const frame = element.ownerDocument.defaultView;
+    for (const ancestor of ancestors(element)) {
+        const isElement = ancestor === element;
+        const style = isElement ? elementStyle : getComputedStyle(ancestor);
+        if (style.opacity === '0') {
+            return false;
+        }
+        if (style.display === 'contents') {
+            // display: contents elements have no box themselves, but children are
+            // still rendered.
+            continue;
+        }
+        const rect = isElement ? elementRect : ancestor.getBoundingClientRect();
+        if ((rect.width === 0 || rect.height === 0) && elementStyle.overflow === 'hidden') {
+            // Zero-sized ancestors don’t make descendants hidden unless the descendant
+            // has overflow: hidden
+            return false;
+        }
+        // Check if the element is off-screen
+        if (isElement && ((rect.right + frame.scrollX < 0) || (rect.bottom + frame.scrollY < 0))) {
+            return false;
+        }
+    }
+    return true;
 }
 
 /**
--- a/utilsForFrontend.mjs
+++ b/utilsForFrontend.mjs
@@ -459,30 +459,29 @@ export function sigmoid(x) {
 
 /**
  * Return whether an element is practically visible, considing things like 0
- * size or opacity, ``display: none``, and ``visibility: hidden``.
+ * size or opacity, and clickability.
  */
 export function isVisible(fnodeOrElement) {
     const element = toDomElement(fnodeOrElement);
-    for (const ancestor of ancestors(element)) {
-        const style = getComputedStyle(ancestor);
-        if (style.visibility === 'hidden' ||
-            style.display === 'none' ||
-            style.opacity === '0' ||
-            style.width === '0px' ||
-            style.height === '0px') {
-            return false;
-        } else {
-            // It wasn't hidden based on a computed style. See if it's
-            // offscreen:
-            const rect = element.getBoundingClientRect();
-            const frame = element.ownerDocument.defaultView;  // window or iframe
-            if ((rect.right + frame.scrollX < 0) ||
-                (rect.bottom + frame.scrollY < 0)) {
-                return false;
-            }
-        }
+    const rect = element.getBoundingClientRect();
+    if (rect.width === 0 || rect.height === 0) {
+        return false;
     }
-    return true;
+    const style = getComputedStyle(element);
+    if (style.opacity === '0') {
+        return false;
+    }
+    // workaround for https://github.com/w3c/csswg-drafts/issues/4122
+    const scrollX = window.pageXOffset;
+    const scrollY = window.pageYOffset;
+    const absX = rect.x + scrollX;
+    const absY = rect.y + scrollY;
+    window.scrollTo(absX, absY);
+    const newX = absX - window.pageXOffset;
+    const newY = absY - window.pageYOffset;
+    const eles = document.elementsFromPoint(newX, newY);
+    window.scrollTo(scrollX, scrollY);
+    return eles.includes(element);
 }
 
 /**
--- a/utilsForFrontend.mjs
+++ b/utilsForFrontend.mjs
@@ -460,11 +460,9 @@ export function sigmoid(x) {
 /**
  * Return whether an element is practically visible, considering things like 0
  * size or opacity, ``visibility: hidden`` and ``overflow: hidden``.
- *
- * This could be 5x more efficient if https://github.com/w3c/csswg-drafts/issues/4122
- * happens.
  */
 export function isVisible(fnodeOrElement) {
+    // This could be 5x more efficient if https://github.com/w3c/csswg-drafts/issues/4122 happens.
     const element = toDomElement(fnodeOrElement);
     // Avoid reading ``display: none`` due to Bug 1381071
     const elementRect = element.getBoundingClientRect();
@@ -475,7 +473,14 @@ export function isVisible(fnodeOrElement) {
     if (elementStyle.visibility === 'hidden') {
         return false;
     }
+    // Check if the element is off-screen
     const frame = element.ownerDocument.defaultView;
+    if (elementRect.x + elementRect.width < 0
+        || (elementRect.y + elementRect.height) < 0
+        || (elementRect.x > frame.innerWidth || elementRect.y > frame.innerHeight)
+    ) {
+        return false;
+    }
     for (const ancestor of ancestors(element)) {
         const isElement = ancestor === element;
         const style = isElement ? elementStyle : getComputedStyle(ancestor);
@@ -491,10 +496,6 @@ export function isVisible(fnodeOrElement) {
         if ((rect.width === 0 || rect.height === 0) && elementStyle.overflow === 'hidden') {
             // Zero-sized ancestors don’t make descendants hidden unless the descendant
             // has overflow: hidden
-            return false;
-        }
-        // Check if the element is off-screen
-        if (isElement && ((rect.right + frame.scrollX < 0) || (rect.bottom + frame.scrollY < 0))) {
             return false;
         }
     }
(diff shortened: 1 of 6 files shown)


danielhertenstein:
Does this mean that rule didn't work exactly properly before?

erikrose:
Yep!

utilsForFrontend.mjs
```diff
@@ -468,8 +468,8 @@ export function isVisible(fnodeOrElement) {
         if (style.visibility === 'hidden' ||
             style.display === 'none' ||
             style.opacity === '0' ||
-            style.width === '0' ||
-            style.height === '0') {
+            style.width === '0px' ||
```
erikrose: While we're Doing Right Things, we might as well also cover cases like "0%", which I just happened to hit while testing this. My gosh, there are going to be a lot of them. :-( Maybe we should just rely on getBoundingClientRect instead? Would that always show a zero height or width when there is one?

biancadanforth: Ah, okay, what if we use a variation on Emilio's recommended version from the [Price Tracker issue](https://github.com/mozilla/price-tracker/issues/319) which omits `getComputedStyle().width/height` entirely, but add the `else` statement you have here regarding offscreen elements?

biancadanforth:
Hm... `isVisible` didn't have any test coverage initially, and I don't see anything set up for functional testing. Perhaps that can be a follow-up issue if desired?

erikrose:
Yes, there's no testing right now that uses an actual browser. It would be great to rig some for error-prone things like this function. Please open an issue!

utilsForFrontend.mjs
erikrose: Sounds good. I assume we'd add something to maintain this offscreen-detection functionality from the current version:

```js
            // It wasn't hidden based on a computed style. See if it's
            // offscreen:
            const rect = element.getBoundingClientRect();
            const frame = element.ownerDocument.defaultView;  // window or iframe
            if ((rect.right + frame.scrollX < 0) ||
                (rect.bottom + frame.scrollY < 0)) {
                return false;
            }
```

biancadanforth: The updated implementation should cover that scenario, as we `window.scrollTo` to update the viewport location before checking for clickability.

I actually ended up going with the third option checking for clickability (see "Updating isVisible: second attempt" in [this comment](https://github.com/mozilla/price-tracker/issues/319#issue-468970781)), but added the workaround that doesn't require privileged Firefox code. This yields a 53% reduction in Fathom-related jank (see the commit message for more details and actual profiles).

If/when the CSS working group updates `element(s)FromPoint` to include the option to ignore viewport clipping, then we can modify this method and we'll see the 81% reduction as described in that comment.

biancadanforth:
@erikrose Ready for your review! As discussed, this patch includes only the correctness fix and a performance improvement that excludes memoization, which is covered in #121 .

utilsForFrontend.mjs
```diff
@@ -459,30 +459,29 @@ export function sigmoid(x) {
...
+    if (style.opacity === '0') {
+        return false;
     }
-    return true;
+    // workaround for https://github.com/w3c/csswg-drafts/issues/4122
+    const scrollX = window.pageXOffset;
+    const scrollY = window.pageYOffset;
+    const absX = rect.x + scrollX;
+    const absY = rect.y + scrollY;
+    window.scrollTo(absX, absY);
```
erikrose: Does this make the page visibly jump? If not now, is there any guaranteed that it won't start visibly jumping in a later Firefox (or other browser)?

utilsForFrontend.mjs
```diff
@@ -459,30 +459,29 @@ export function sigmoid(x) {
...
+    const newX = absX - window.pageXOffset;
+    const newY = absY - window.pageYOffset;
+    const eles = document.elementsFromPoint(newX, newY);
```
erikrose: It's a nit, but I'd go with the full `elements` or something else unabbreviated here. We don't really have a convention of abbreviating things in this codebase (apart from fnodes, I suppose, but that has a glossary entry of its own).

utilsForFrontend.mjs
biancadanforth: Admittedly, this workaround is awkward. Though I got some heartening information from Emilio/the web[1], if the user is scrolling while `isVisible` happens to be running, there can be jumpiness. I'll have to revise my approach here.

[1]:
1. I asked Emilio, the graphics/layout engineer who has been helping me analyze alternative approaches:

> I think that work around is ok.

> Javascript needs to run before rendering, the engine cannot simply
interrupt a script and do layout in order to figure out what changed.

> Those calls are sync so I think you're ok.

> My only concern is what happens if the user tries to scroll while that
code is running, since scrolling _is_ async. But I think it should be
fine since we only tell the compositor about the main thread scroll
state after painting.

2. I was reading a [document on best practices](https://developers.google.com/web/fundamentals/performance/rendering/optimize-javascript-execution#reduce_complexity_or_use_web_workers), and it recommends:
> ...if you’re in an animation like scrolling, you should ideally be looking to keep your JavaScript to something in the region of 3-4 ms.

When I look at this implementation’s [profile](https://perfht.ml/2K7Vqhs) in Price Tracker (the Stack Chart, Focus: isVisible), I don’t see any calls exceeding 2 ms, and the average looks to be closer to 1 ms.

Screen Shot 2019-07-31 at 11 34 04 AM

erikrose:
In today's team meeting, we decided that scrollTo bore too much risk of making UI seem wonky. We decided to land the implementation that makes privileged code very fast while containing an `else` clause that makes non-privileged code merely correct (which the existing implementation isn't) and somewhat fast.

biancadanforth:
@erikrose , would you like me to submit the Firefox-specific solution as part of this PR or in a separate PR?

erikrose:
Let's do a separate one. I almost always lean toward getting *something* landed earlier so people can start enjoying it—and in case we get interrupted. Thanks! :-D

biancadanforth:
Okay thanks @erikrose ; I agree with you on that. It's good to keep PRs small. This will make #122 more reasonable as well. I have requested your review then.

utilsForFrontend.mjs
```diff
@@ -458,28 +458,44 @@ export function sigmoid(x) {
...
+            continue;
+        }
+        const rect = isElement ? elementRect : ancestor.getBoundingClientRect();
+        if ((rect.width === 0 || rect.height === 0) && elementStyle.overflow === 'hidden') {
+            // Zero-sized ancestors don’t make descendants hidden unless the descendant
+            // has overflow: hidden
+            return false;
+        }
+        // Check if the element is off-screen
+        if (isElement && ((rect.right + frame.scrollX < 0) || (rect.bottom + frame.scrollY < 0))) {
```
erikrose: Unless I'm mistaken, you can move this to before the `for` loop, strip off the `isElement` condition, and it'll be equivalent.

biancadanforth: Yes, I can do that. I also modified the check, as previously, I hadn't checked all four conditions (element is above, to the left, to the right or below the visible portion of the page).

biancadanforth:
@erikrose , This is ready for another look!

erikrose:
Oh noes! Since you force-pushed, now I don't know what's new anymore. :-( Let me see if I can figure it out.

biancadanforth:
@erikrose Just the last two commits are new.

erikrose:
Thanks. Just as soon as I said that and reloaded the page, GitHub started showing me "These commits are new since you reviewed". :-)

biancadanforth:
Yeah I was just editing the commit subject from one of the earlier commits to be more accurate. I'll try to keep rebasing to a minimum to avoid confusion.

utilsForFrontend.mjs
```diff
@@ -475,7 +473,14 @@ export function isVisible(fnodeOrElement) {
     if (elementStyle.visibility === 'hidden') {
         return false;
     }
+    // Check if the element is off-screen
```
erikrose: I was saying to Daniel the other day that I like to punctuate comments like sentences if they are, in fact, sentences. I think it's influence from https://www.python.org/dev/peps/pep-0008/#comments. But on comments like this that take up the whole line and refer to code below, I often stick a colon afterward. That way, even if you don't skip a line before, it's clear you're talking about code only below it.

utilsForFrontend.mjs
```diff
@@ -475,7 +473,14 @@ export function isVisible(fnodeOrElement) {
...
     const frame = element.ownerDocument.defaultView;
+    if (elementRect.x + elementRect.width < 0
+        || (elementRect.y + elementRect.height) < 0
+        || (elementRect.x > frame.innerWidth || elementRect.y > frame.innerHeight)
```
erikrose: It seems like the parens around this pair of conditions shouldn't be here, since you didn't put them around the top pair (and they don't affect the logic).

utilsForFrontend.mjs
```diff
@@ -475,7 +473,14 @@ export function isVisible(fnodeOrElement) {
     if (elementStyle.visibility === 'hidden') {
         return false;
     }
+    // Check if the element is off-screen
     const frame = element.ownerDocument.defaultView;
+    if (elementRect.x + elementRect.width < 0
+        || (elementRect.y + elementRect.height) < 0
```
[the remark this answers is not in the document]
biancadanforth: Do we want to add an eslint rule on parens? e.g. `[no-extra-parens](https://eslint.org/docs/rules/no-extra-parens): error` I have no strong opinion. I don't know that there is one to enforce use of extra parens.

It sounds like you do have an opinion about operators in conditional statements, so I added an eslint rule for that in `.eslintrc.yml`: `operator-linebreak: [error, after]` and fixed the one other instance of this elsewhere in `utilsForFrontend.mjs`.

utilsForFrontend.mjs
biancadanforth: I can get behind the complete sentence thing. Not sure if I will remember your preference around colons, but I'll try to keep that in mind...I added a colon.

biancadanforth:
@erikrose ready for another pass!

utilsForFrontend.mjs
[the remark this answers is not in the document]
erikrose: I don't see a benefit in making everybody memorize the precedence of every little operator, which would seem to be a requirement for no-extra-parens. I'm just looking for consistency. When I see inconsistency, I think "Aha, the author is trying to show me that something is different about this other case," and I waste time looking for it. That's all.

utilsForFrontend.mjs
```diff
@@ -160,8 +160,8 @@ export function *inlineTexts(element, shouldTraverse = element => true) {
     for (let child of walk(element,
                            element => !(isBlock(element) ||
                                         element.tagName === 'SCRIPT' &&
-                                        element.tagName === 'STYLE')
-                                      && shouldTraverse(element))) {
+                                        element.tagName === 'STYLE') &&
```
erikrose: I think this actually make it harder to understand. I don't remember writing this (though I probably did), but here's how I read the original:

The ! lines up with the &&, suggesting that 2 clauses are ANDed together. That's what's actually going on.

When you move the 2nd &&, the 2 &&s line up exactly, and it starts looking like shouldTraverse is just one more clause in the earlier &&, unless you take the time to painstakingly balance the parens.

So I think the original indentation visually conveys the structure more clearly. Does that make sense?

utilsForFrontend.mjs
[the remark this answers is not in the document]
erikrose: As for the operator placement, I was just trying to make the very similar lines line up for faster scanning.

utilsForFrontend.mjs
```diff
@@ -464,20 +464,20 @@ export function sigmoid(x) {
 export function isVisible(fnodeOrElement) {
     // This could be 5x more efficient if https://github.com/w3c/csswg-drafts/issues/4122 happens.
     const element = toDomElement(fnodeOrElement);
-    // Avoid reading ``display: none`` due to Bug 1381071
     const elementRect = element.getBoundingClientRect();
-    if (elementRect.width === 0 && elementRect.height === 0) {
+    const elementStyle = getComputedStyle(element);
+    // Alternative to reading ``display: none`` due to Bug 1381071.
+    if (elementRect.width === 0 && elementRect.height === 0 && elementStyle.overflow !== 'hidden') {
```
erikrose: Shouldn't this be `elementStyle.overflow === 'hidden'`?  As you say below, "Zero-sized ancestors don’t make descendants hidden unless the descendant has overflow: hidden".

utilsForFrontend.mjs
biancadanforth: I see what you mean, though that prevents us from having an eslint rule to cover it, as leaving this as-is is not consistent (from eslint's perspective) with changing [the other one](https://github.com/mozilla/fathom/pull/116#discussion_r312537506).

I'm happy to change it back as you prefer, however.

utilsForFrontend.mjs
biancadanforth: The check you're thinking of is done inside the `for` loop below. This first check is a proxy for checking `display: none`, for which just checking `width` and `height` === 0 is not sufficient as I learned while working on #122 .

biancadanforth:
@erikrose , Ready for another look.

utilsForFrontend.mjs
erikrose: Ah, so this is a check strictly for the visibility of the element passed in, not any of its descendents. A child, for example, could still be visible. Callers will have to be aware of this. That'd be worth documenting in the doclet, as I wasn't front-end-savvy enough to realize it.

biancadanforth: Hm... I'm not sure I follow what you want me to document. In general, this method is to check if the element passed in is visible. It doesn't make any claims about that element's descendants.

biancadanforth: We discussed this synchronously, and we're just going to leave off any additional comments for now.

PR #117, opened 2019-07-19
Adds fathom-serve tool

danielhertenstein:
This PR adds another `fathom-` tool! This time to run an https server for you for serving pages for the vectorizer in FathomFox. 

We are in the process of extracting resources out of our freezeDry'd pages (see #102) and after extracting the resources, many pages will not load their resources and will complain with the following message in the developer console:

`Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at file:///C:/blahblahblah/resources/fp34_resources/6.woff2. (Reason: CORS request not http).`

See [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSRequestNotHttp) for more details.

I have one question and one potential problem to bring to your attention:
- Question: Is there any reason we should run the server off in a different thread allowing the user to close the server with a keypress rather than blocking the thread and forcing the user to stop the script with a `crtl+c`?
- Problem: The certificate in this PR is only a self-signed certificate, which the browser may complain about. The browser did **not** complain when I did a run of vectorization on my machine in a fresh instance of my dev profile of nightly, but I would like a second tester.


erikrose:
I don't see a problem with ctrl-C, especially if we print "Hey, press control-C to stop."

cli/fathom_web/commands/serve.py
```diff
@@ -0,0 +1,36 @@
...
+    handler = partial(SimpleHTTPRequestHandler, directory=directory)
+    server = HTTPServer(('localhost', port), handler)
+    certfile = pathlib.Path(os.path.realpath(__file__)).parent / '..' / 'cert' / 'cert.pem'
+    server.socket = ssl.wrap_socket(
+        server.socket,
+        certfile=certfile,
+        server_side=True,
+    )
+    print(f'Serving {directory} over https://localhost:{port}')
+    server.serve_forever()
```
erikrose: And let's add the "Press control-C to stop." thing

erikrose:
> The certificate in this PR is only a self-signed certificate, which the browser may complain about

It complains the first time I load a page from it, and then it's quiet after that.

cli/fathom_web/cert/cert.pem
```diff
@@ -0,0 +1,50 @@
+-----BEGIN PRIVATE KEY-----
```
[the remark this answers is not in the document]
danielhertenstein: 10 years

cli/fathom_web/commands/serve.py
```diff
@@ -0,0 +1,36 @@
...
+        help='The port to use (default: 8000)')
+@argument('directory', type=Path(exists=True, file_okay=False))
+def main(directory, port):
+    """
+    Serves the files in DIRECTORY over a local https server: https://localhost:PORT.
+
+    The default PORT is 8000.
+
+    This is useful for vectorizing samples using FathomFox. FathomFox expects you to provide,
+    in the vectorizer page, an address to an https server that is serving your samples.
```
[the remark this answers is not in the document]
danielhertenstein: Will do!

cli/fathom_web/commands/serve.py
```diff
@@ -0,0 +1,36 @@
...
+import pathlib
+import ssl
+
+from click import argument, command, option, Path
+
+
+@command()
+@option('--port', '-p', type=int, default=8000,
+        help='The port to use (default: 8000)')
+@argument('directory', type=Path(exists=True, file_okay=False))
```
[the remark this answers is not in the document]
danielhertenstein: I like this. Coming it next set of commits.

cli/fathom_web/commands/serve.py
danielhertenstein: Will do!

danielhertenstein:
> 
> 
> > The certificate in this PR is only a self-signed certificate, which the browser may complain about
> 
> It complains the first time I load a page from it, and then it's quiet after that.

Do you think that is too annoying for users? The alternative is to get a more legitimate certificate that we have to worry about updating, right?

erikrose:
Let's Encrypt has [a good article](https://letsencrypt.org/docs/certificates-for-localhost/) about the possibilities. I think what you have here is about the best you can get. The one other thing that might be worth a shot is issuing a cert with the CN of 127.0.0.1 and seeing if things work better when you put https://127.0.0.1 in the browser. The article talks about how behavior is subtly different with the IP.

cli/fathom_web/commands/serve.py
```diff
@@ -0,0 +1,36 @@
...
+    Serves the files in DIRECTORY over a local https server: https://localhost:PORT.
+
+    The default PORT is 8000.
+
+    This is useful for vectorizing samples using FathomFox. FathomFox expects you to provide,
+    in the vectorizer page, an address to an https server that is serving your samples.
+    """
+    handler = partial(SimpleHTTPRequestHandler, directory=directory)
+    server = HTTPServer(('localhost', port), handler)
+    certfile = pathlib.Path(os.path.realpath(__file__)).parent / '..' / 'cert' / 'cert.pem'
```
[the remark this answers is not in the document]
danielhertenstein: Done!

danielhertenstein:
Interesting. I'll take a look and try it out.

danielhertenstein:
Using 127.0.0.1 as the common name didn't change the warning the browser initially gives.

erikrose:
Thanks for checking.

cli/fathom_web/commands/serve.py
```diff
@@ -4,18 +4,20 @@
...
 @option('--port', '-p', type=int, default=8000,
         help='The port to use (default: 8000)')
-@argument('directory', type=Path(exists=True, file_okay=False))
+@option('--directory', '-d', type=Path(exists=True, file_okay=False), default=os.getcwd(),
+        help='The directory to serve files from (default: current working directory')
 def main(directory, port):
     """
     Serves the files in DIRECTORY over a local https server: https://localhost:PORT.
 
     The default PORT is 8000.
```
erikrose: I don't think we need to repeat the defaults here (where they will probably fall out of date anyway) when `--help` auto-formats them from the help strings.

cli/fathom_web/commands/serve.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    handler = partial(SimpleHTTPRequestHandler, directory=directory)
+    server = HTTPServer(('localhost', port), handler)
+    certfile = resource_filename('fathom_web', 'cert/cert.pem')
+    server.socket = ssl.wrap_socket(
+        server.socket,
+        certfile=certfile,
+        server_side=True,
+    )
+    print(f'Serving {directory} over https://localhost:{port}')
```
erikrose: Periods after the sentences

cli/fathom_web/cert/cert.pem
erikrose: I feel like we might as well move the single file out of the dir and let it be top level. No?

cli/fathom_web/commands/serve.py
danielhertenstein: That is a good point.

cli/fathom_web/commands/serve.py
danielhertenstein: You got it!

cli/fathom_web/cert/cert.pem
danielhertenstein: I do think having a directory named `cert` is a bit much. While we just have this single file, I think letting it be top level is fine. If we start having more data files, perhaps we'll want a `data` directory.

biancadanforth:
So this new `fathom-serve` tool solves the CORS issue by providing a CSP for the frozen page that allows COR loading as long as the page itself is the one requesting the resource?

Why is the certificate encrypted? When I was reading about [CSPs on MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP), these directives seem to be in headers in plain text.

If/when we move to remote hosting of pages, we could probably talk to Christoph K. He works on content security.

danielhertenstein:
@biancadanforth, this tool solve the CORS issue by allowing the user to use HTTPS URLs when requesting resources (see [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSRequestNotHttp)). The previous way of serving the images for the vectorizer was with an HTTP server and therefore HTTP URLs.

The CSP problem I ran into is completely separate (though I did not see it until the CORS issue was solved), and is solved in #109 by adding to the directives in the header that freezeDry puts in place.

biancadanforth:
Ah okay, thanks @danielhertenstein , so the cert file in this PR is what enables us to use HTTPS instead of HTTP?

danielhertenstein:
Yup!

Issue #125, opened 2019-07-30
`fathom-pick` needs to move sub-resource directories


danielhertenstein:
Now that `fathom-extract` exists, samples are more than just their HTML page, they have a directory of resources that should be in a resource directory that is on the same level as the HTML page. For page rendering to work properly, the relative path needs to hold. When someone uses `fathom-pick` to organize samples, the resource directories need to be moved appropriately.

Add logic to `fathom-pick` such that for each HTML page that will be moved, the presence of a page-specific resource directory should be checked for in `resources/`. If it does exist, we should move that page's resource directory to the destination directory of `fathom-pick`. If there is not yet a `resources/` directory at the destination, it should be made. If there already exists a page-specific resource directory in the destination `resources/` directory, raise an exception instead of overwriting the files.

Because I've used the word resources too many times above for my own liking here is an example:

`fathom-pick` source directory:
```
samples/
    resources/
        1/
        2/
    1.html
    2.html
    3.html
```
`fathom-pick` destination directory:
```
training/
```
When someone runs `fathom-pick samples training 2`, let's pretend files `1.html` and `3.html` are picked. In this case, `fathom-pick` will see that there is a resources directory for `1.html` and will create a new `resources/1/` directory in `training/`. Since there are no resources for `3.html` there is no additional filesystem work beyond moving `3.html`. The resulting directories should now look like:
```
samples/
    resources/
        2/
    2.html

training/
    resources/
        1/
    1.html
    3.html
```

danielhertenstein:
Closed by #126

PR #126, opened 2019-08-05
Move resources with pick

danielhertenstein:
Addresses #125 

This PR changes `fathom-pick` such that it specifically moves HTML files and any extracted resources. Now that samples will almost certainly have an associated directory of extracted resources, we need to move those resources along with the HTML files when dividing samples into training/validation/test so the references to the resources in the HTML still work.

### DIFF
--- a/cli/fathom_web/commands/pick.py
+++ b/cli/fathom_web/commands/pick.py
@@ -2,7 +2,7 @@ import pathlib
 from random import sample
 from shutil import move
 
-from click import argument, command, Path
+from click import argument, command, Path, UsageError
 
 
 @command()
@@ -23,14 +23,13 @@ def main(from_dir, to_dir, number):
     from_dir = pathlib.Path(from_dir)
     to_dir = pathlib.Path(to_dir)
 
-    files = [f for f in from_dir.glob('*.html')]
-    for file in sample(files, number):
-        # If the file has resources, we must move those as well
+    for file in sample(list(from_dir.glob('*.html')), number):
+        # If the file has resources, we must move those as well:
         if (from_dir / 'resources' / file.stem).exists():
             # Make sure we don't overwrite an existing resources directory
             if (to_dir / 'resources' / file.stem).exists():
-                raise RuntimeError(f'Tried to make directory {(to_dir / "resources" / file.stem).as_posix()}, but it'
-                                   f' already exists. To protect against unwanted data loss, please move or remove the'
-                                   f' existing directory.')
+                raise UsageError(f'Tried to make directory {(to_dir / "resources" / file.stem).as_posix()}, but it'
+                                 f' already exists. To protect against unwanted data loss, please move or remove the'
+                                 f' existing directory.')
             move(from_dir / 'resources' / file.stem, to_dir / 'resources' / file.stem)
         move(file.as_posix(), to_dir)
--- a/cli/fathom_web/commands/pick.py
+++ b/cli/fathom_web/commands/pick.py
@@ -12,7 +12,7 @@ from click import argument, command, Path
           type=Path(exists=True, file_okay=False, writable=True, dir_okay=True, allow_dash=True))
 @argument('number', type=int)
 def main(from_dir, to_dir, number):
-    """Move a given number of html files and any extracted resources from one directory to another.
+    """Move a given number of HTML files and any extracted resources from one directory to another.
 
     Ignore hidden files.
 
--- a/cli/fathom_web/commands/pick.py
+++ b/cli/fathom_web/commands/pick.py
@@ -1,5 +1,4 @@
-from os import listdir
-from os.path import join
+import pathlib
 from random import sample
 from shutil import move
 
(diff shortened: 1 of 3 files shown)


cli/fathom_web/test/test_pick.py
```diff
@@ -0,0 +1,63 @@
...
+
+    # Check the correct number of files have moved
+    files_in_source = list(source.glob('*.html'))
+    assert len(files_in_source) == 1
+    files_in_destination = list(destination.glob('*.html'))
+    assert len(files_in_destination) == 2
+
+    # Check any resource directories have moved
+    if (destination / '1.html').exists():
+        assert (destination / 'resources' / '1' / '1.png').exists()
```
erikrose: I like this inlining and would also prefer it for the setup code (`file_1_resources = ...`), especially since we never actually read half those variables. Fewer var names blowing around make it easier for me to understand the code, since it's less state to track.

cli/fathom_web/test/test_pick.py
```diff
@@ -61,3 +61,50 @@ def test_end_to_end(tmp_path):
...
+    file_1 = source / '1.html'
+    file_1.write_text('I am file 1')
+
+    # Add the resource directory for our file
+    resources = source / 'resources'
+    resources.mkdir()
+    file_1_resources = resources / '1'
+    file_1_resources.mkdir()
+    file_1_resource_1 = file_1_resources / '1.png'
+    file_1_resource_1.write_text('I am resource 1 for file 1')
```
erikrose: Since we aren't asserting anything about the file contents, does it do us any good to write any?

cli/fathom_web/commands/pick.py
```diff
@@ -13,14 +12,25 @@
...
 
     Ignore hidden files.
 
     This is useful for dividing a corpus into training, validation, and testing
     parts.
-
     """
-    files = [f for f in listdir(from_dir) if not f.startswith('.')]
+    # Make these strings into ``Path``s so they are easier to work with
+    from_dir = pathlib.Path(from_dir)
```
erikrose: Any reason not to import more deeply and save some reading?

cli/fathom_web/commands/pick.py
```diff
@@ -13,14 +12,25 @@
...
+    to_dir = pathlib.Path(to_dir)
+
+    files = [f for f in from_dir.glob('*.html')]
```
erikrose: I might have said `list(from_dir...)` just for brevity, since you aren't doing any transformation of the values. Opinions?

cli/fathom_web/commands/pick.py
```diff
@@ -13,14 +12,25 @@
...
     for file in sample(files, number):
-        move(join(from_dir, file), to_dir)
+        # If the file has resources, we must move those as well
+        if (from_dir / 'resources' / file.stem).exists():
```
erikrose: Darn but this Path stuff reads well! :-D

cli/fathom_web/commands/pick.py
```diff
@@ -13,14 +12,25 @@
...
     """
-    files = [f for f in listdir(from_dir) if not f.startswith('.')]
+    # Make these strings into ``Path``s so they are easier to work with
+    from_dir = pathlib.Path(from_dir)
+    to_dir = pathlib.Path(to_dir)
+
+    files = [f for f in from_dir.glob('*.html')]
     for file in sample(files, number):
-        move(join(from_dir, file), to_dir)
+        # If the file has resources, we must move those as well
```
erikrose: *puts on his annoying "Sentence comments should be punctuated like sentences" hat*

That said, I often use colons instead of periods if I'm referring to the following line. It makes it clear that it isn't talking about code above.

cli/fathom_web/test/test_pick.py
danielhertenstein: That is a good point. I have those changes queued up and there are less lines as a result.

cli/fathom_web/test/test_pick.py
danielhertenstein: It does us no good. I forgot about the method `Path.touch()` to make empty files. I have that change queued up.

cli/fathom_web/commands/pick.py
danielhertenstein: `click` also has a `Path` class, so I can't import `Path` here.

cli/fathom_web/commands/pick.py
danielhertenstein: I like the colon style. I'll make that change.

cli/fathom_web/commands/pick.py
danielhertenstein: Yea, this line is awful. Thank you for catching it. I've queued up a change to get rid of it and use `list(from_dir...)` directly in the call to `sample()`.

cli/fathom_web/commands/pick.py
erikrose: Ah yes. Bummer. I'd deep-import whichever one I use more often.

cli/fathom_web/commands/pick.py
erikrose: Heh, I'd stop short of "awful". ;-)

cli/fathom_web/commands/pick.py
```diff
@@ -13,14 +12,25 @@
...
+        if (from_dir / 'resources' / file.stem).exists():
+            # Make sure we don't overwrite an existing resources directory
+            if (to_dir / 'resources' / file.stem).exists():
+                raise RuntimeError(f'Tried to make directory {(to_dir / "resources" / file.stem).as_posix()}, but it'
```
erikrose: It seems to me we should raise click.Abort instead: https://click.palletsprojects.com/en/7.x/exceptions/#which-exceptions-exist. It's a teeny bit more descriptive, and it probably results in a more nicely displayed error.

cli/fathom_web/test/test_pick.py
erikrose: Cool. I'll wait to see those before I say anything else about the tests, which is likely to take the form of "factor out repetition" if anything.

erikrose:
I'm not sure what's going on with the JS tests on Travis, but I doubt you broke them here. [Time passes.] Okay, I restarted the Travis job, and it worked this time. Maybe some APT repo was down. I do see some places where `out()` has become an unused import due to your `out`-removal work:

```
home/travis/build/mozilla/fathom/test/demos.mjs

  3:14  warning  'out' is defined but never used  no-unused-vars

/home/travis/build/mozilla/fathom/test/lhs_tests.mjs

  3:29  warning  'out' is defined but never used  no-unused-vars

/home/travis/build/mozilla/fathom/test/ruleset_tests.mjs

  4:28  warning  'out' is defined but never used  no-unused-vars

/home/travis/build/mozilla/fathom/test/utils_tests.mjs

  2:14  warning  'out' is defined but never used  no-unused-vars
```

We should fix those in another PR sometime.

danielhertenstein:
I'm happy to fix those warnings with a commit to master. It should be as simple as removing the imports in those files. I definitely caused that with my `out()` PR.

erikrose:
Look like Bianca already did it in one of her soon-to-land PRs! :-D

danielhertenstein:
Nice!

Issue #123, opened 2019-07-29
Add fathom-extract regex tests


danielhertenstein:
The regex needs tests. I will add these when #120 is through.

PR #127, opened 2019-08-09
Adds regex tests for fathom-extract

danielhertenstein:
Addresses #123.

This PR adds the cases I could remember. These tests will help protect against regressions whenever the regex needs to change.

### DIFF
--- a/cli/fathom_web/test/test_extract.py
+++ b/cli/fathom_web/test/test_extract.py
@@ -1,0 +1,32 @@
+from ..commands.extract import BASE64_DATA_PATTERN
+
+
+def test_common_example():
+    """Confirm we handle to well behaving case"""
+    mime_type = 'image/png'
+    base64_string = 'aorienstar/tar/ararnsoine98daQAAAIST+++/rstienf='
+    test_string = f'data:{mime_type};base64,{base64_string}'
+    matches = list(BASE64_DATA_PATTERN.finditer(test_string))
+    assert len(matches) == 1
+    assert matches[0].group('mime') == mime_type
+    assert matches[0].group('string') == base64_string
+
+
+def test_empty_string():
+    """Some base64 strings are actually empty"""
+    test_string = 'data:;base64,'
+    matches = list(BASE64_DATA_PATTERN.finditer(test_string))
+    assert len(matches) == 0
+
+
+def test_presence_of_charset():
+    """Some base64 strings contain a character set specification"""
+    test_string = 'data:image/png; charset=utf-8;base64,iVBORw0K'
+    matches = list(BASE64_DATA_PATTERN.finditer(test_string))
+    assert len(matches) == 1
+
+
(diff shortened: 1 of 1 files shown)


cli/fathom_web/test/test_extract.py
```diff
@@ -0,0 +1,32 @@
+from ..commands.extract import BASE64_DATA_PATTERN
+
+
+def test_common_example():
+    """Confirm we handle to well behaving case"""
```
erikrose: to→the

And it's a sentence, so period! :-) If you want to get fancy, you can hyphenate "well-behaving". :-o

cli/fathom_web/test/test_extract.py
```diff
@@ -0,0 +1,32 @@
...
+    mime_type = 'image/png'
+    base64_string = 'aorienstar/tar/ararnsoine98daQAAAIST+++/rstienf='
+    test_string = f'data:{mime_type};base64,{base64_string}'
+    matches = list(BASE64_DATA_PATTERN.finditer(test_string))
```
erikrose: Can we call findall() instead and save a list() call?

cli/fathom_web/test/test_extract.py
```diff
@@ -0,0 +1,32 @@
...
+    assert len(matches) == 1
+    assert matches[0].group('mime') == mime_type
```
erikrose: I really like that you factored out the MIME type and such.

cli/fathom_web/test/test_extract.py
```diff
@@ -0,0 +1,32 @@
...
+    assert matches[0].group('string') == base64_string
+
+
+def test_empty_string():
+    """Some base64 strings are actually empty"""
+    test_string = 'data:;base64,'
+    matches = list(BASE64_DATA_PATTERN.finditer(test_string))
```
erikrose: This line would probably be a good thing to factor out into a test-helper procedure. It's used in every test, it's got a few different pieces of knowledge in it, and I already requested one tweak to it.

cli/fathom_web/test/test_extract.py
danielhertenstein: Whoops. Thanks for catching that!

cli/fathom_web/test/test_extract.py
danielhertenstein: `findall()` does not return `Match` objects, while `finditer()` does. Since we use the `Match` objects in `fathom-extract`, we should test with `Match` objects as well.

cli/fathom_web/test/test_extract.py
danielhertenstein: That's reasonable. I'm on it.

cli/fathom_web/test/test_extract.py
erikrose: Good point! I didn't see that in my quick read.

Issue #122, opened 2019-07-29
Add JavaScript functional test harness


biancadanforth:
As discussed [here](https://github.com/mozilla/fathom/pull/116#issuecomment-515270606), there is no test coverage for functions exported by Fathom's JavaScript library that require rendering a page in a browser.

Given that we've already had issues with some of these error-prone functions (like `isVisible`, see #118 ), it would be wise to set up some automated testing for them in our CI.

As Erik mentioned, we may be able to achieve this using [Firefox's Headless mode](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Headless_mode).

As part of closing this issue, a test for `isVisible` should be written.

If we prioritize this issue, I am very interested in working on it. For now, I will leave it unassigned.

biancadanforth:
I spoke with jlast who is on the DevTools team and in part works on their CI.

For a web development test harness (no compatibility with landing code in Firefox), he recommends [Cypress](https://www.cypress.io/), which is [Electron](https://electronjs.org/) based. DevTools used this themselves until they switched to [Mochitest](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Mochitest) (Firefox only).

There is ongoing work to make significant improvements to [Puppeteer](https://pptr.dev/) for Firefox, which he thinks will be complete around December 2019. At that point, he would recommend Puppeteer.

erikrose:
I watched Cypress's intro movie, and my takeaways (mostly for my own reference) were…
1. It looks amazing UI-wise for building tests of complicated, multi-step interactions with JS apps. Our use case, however, is going to be dominated by unit-like tests of utilsForFrontEnd.js helpers for the forseeable future.
2. They're pushing a subscription-based payment plan for their own Travis-like web-based frontend. It's not immediately obvious how to use it on our existing infra, but I don't know that it's impossible.
3. It's monolithic, including (and requiring?) its own assertion lib and test runner.

erikrose:
A couple more links to follow (sorry, I got interested and couldn't help it):
* https://docs.travis-ci.com/user/gui-and-headless-browsers/
* https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Headless_mode (has some webdriver stuff to evaluate)

biancadanforth:
> 2. They're pushing a subscription-based payment plan for their own Travis-like web-based frontend. It's not immediately obvious how to use it on our existing infra, but I don't know that it's impossible.

[...]

PR #129, opened 2019-08-10
#122: Add JavaScript functional test harness

biancadanforth:
Edit: This PR is branched off of PR #116 , since part of the issue is to write a test for `isVisible`. It should not be merged until that PR is merged. Only the last 6 commits are new.

#### Open questions:
- I couldn’t figure out how to avoid using `timeout` in mocha in the test, as it exceeds the default 2s allowed. Using it works, but I'm open to alternatives.
- Everything's working except the `istanbul` coverage.
  - If I replace `node_modules/.bin/_mocha` with `node_modules/.bin/mocha` in the Makefile's `coverage` command, my test passes but coverage fails (current state).[1] If I leave `node_modules/.bin/_mocha` as-is, my test fails, but coverage works.[2]
  - I have no idea why these would behave differently or create this trade-off. I think the best course of action may be to update `istanbul`, since we are using a [deprecated version](https://github.com/gotwarlost/istanbul/#istanbul---a-js-code-coverage-tool-written-in-js) (strangely, that repo says it's at version 0.4.5, not 1.1.0..., but my main point is that there's a v2). @erikrose , @danielhertenstein , would either of you have any ideas/suggestions?

[1]: _TravisCI log with `node_modules/.bin/mocha` in `make coverage`:_
```
No coverage information was collected, exit without writing coverage information
```
I found this [related issue](https://github.com/gotwarlost/istanbul/issues/262), which is probably why `_mocha` was there in the first place.

[2]: _TravisCI log (error) with `node_modules/.bin/_mocha` in `make coverage`:_
```
  1 failing

  1) isVisible unprivileged

       Should return false when an element is hidden:

     JavascriptError: ReferenceError: cov_1nb0mc0zi8 is not defined
      at Object.throwDecodedError (node_modules/selenium-webdriver/lib/error.js:550:15)
      at parseHttpResponse (node_modules/selenium-webdriver/lib/http.js:560:13)
      at Executor.execute (node_modules/selenium-webdriver/lib/http.js:486:26)
      at processTicksAndRejections (internal/process/task_queues.js:85:5)
      at async thenableWebDriverProxy.execute (node_modules/selenium-webdriver/lib/webdriver.js:699:17)
      at async Context.<anonymous> (test/functional/isVisible.js:41:38)
```
I tried the `/* istanbul ignore next */` solution suggested [here](https://github.com/istanbuljs/nyc/issues/514#issuecomment-322784724), but my test continued to fail no matter where I put it.


test/functional/setup_http_server
```diff
@@ -0,0 +1,24 @@
...
+
+const PORT = 8000;
+
+const server = http.createServer((request, response) => {
+    const path = url.parse(request.url).pathname;
+    fs.readFile(__dirname + path, 'utf8', (error, data) => {
+        if (error) {
+            response.writeHead(404);
+            response.write(`There was an error: ${error.errno}, ${error.code}`);
+            response.end();
```
erikrose: Looks like you could factor this out.

test/functional/isVisible.html
```diff
@@ -0,0 +1,87 @@
...
+      }
+      #not-visible-12 {
+        bottom: calc(100vh + 2 * var(--size));
+      }
+
+    </style>
+  </head>
+  <body>
+    <h1>isVisible functional test</h1>
+    <div id="not-visible-1"></div>
```
erikrose: I think we could reorganize this to make it more comprehensible.

When I first first encounter each tag, I have to look upward to find out what the corresponding style info actually is. Does choosing not to inline the styles actually give us anything? I don't think we are reusing any of the numbered ones. I do like that you factored out `.off-screen`.

Also, it would be great to have some comments, descriptive names, or something for some of the less obvious ones, like the various `calc`s.

test/functional/isVisible.js
```diff
@@ -0,0 +1,58 @@
+const { assert } = require('chai');
+const firefox = require('selenium-webdriver/firefox');
+const webdriver = require('selenium-webdriver');
+const { ancestors, isDomElement, isVisible, toDomElement } = require('../../utilsForFrontend');
```
erikrose: The rest of the codebase has no spaces inside the braces. That would be a good linter rule.

test/functional/isVisible.js
```diff
@@ -0,0 +1,58 @@
...
+describe('isVisible', () => {
+    const options = new firefox.Options();
+    options.headless();
+
+    const driver = new webdriver.Builder()
+        .forBrowser('firefox')
+        .setFirefoxOptions(options)
+        .build();
+
+    it('Should return false when an element is hidden', async () => {
```
erikrose: So far, the rest of the codebase starts `it(` operands with a lowercase letter. Those all show up in the test output, so they're pretty conspicuous.

test/functional/isVisible.js
```diff
@@ -0,0 +1,58 @@
...
+        const hiddenElementIds = [
+            'not-visible-1',
```
erikrose: Might as well just use a counter rather than error-prone-ly typing all these out, no?

test/functional/isVisible.js
```diff
@@ -0,0 +1,58 @@
...
+            'not-visible-6',
+            'not-visible-7',
+            'not-visible-8',
+            'not-visible-9',
+            'not-visible-10',
+            'not-visible-11',
+            'not-visible-12',
+        ];
+        await driver.get(TEST_PAGE_URL);
+        // await driver.wait(async () => {
```
erikrose: Still need these comments?

test/functional/isVisible.js
```diff
@@ -0,0 +1,58 @@
+const { assert } = require('chai');
+const firefox = require('selenium-webdriver/firefox');
+const webdriver = require('selenium-webdriver');
+const { ancestors, isDomElement, isVisible, toDomElement } = require('../../utilsForFrontend');
+
+const { until, By } = webdriver;
```
erikrose: Any reason you chose not to do just `const {Builder, By, until} = require('selenium-webdriver');` and save a line? Did you especially like the explicitness of `webdriver.Builder` below?

test/functional/isVisible.js
```diff
@@ -0,0 +1,58 @@
...
+        await driver.get(TEST_PAGE_URL);
+        // await driver.wait(async () => {
+        //     const readyState = await driver.executeScript('return document.readyState');
+        //     return readyState === 'complete';
+        // }, WAIT_MS, `Page did not finish loading after ${WAIT_MS} ms`);
+
+        for (const id of hiddenElementIds) {
+            await driver.wait(until.elementLocated(By.id(id)), WAIT_MS);
+            const isElementVisible = await driver.executeScript(`
+                ${ancestors}
```
erikrose: Now *this* is a cool trick.

erikrose: JavaScript™: Code-wheelies as a best practice. ;-)

test/functional/isVisible.js
```diff
@@ -0,0 +1,58 @@
...
+                ${isDomElement}
+                ${toDomElement}
+                return ${isVisible}(document.getElementById('${id}'));
+            `);
+            assert.equal(
+                isElementVisible,
+                false,
+                `isVisible should return false for element with id '${id}'.`
```
erikrose: We should also test at least one case in which isVisible returns true; right now, `function() {return false;}` would pass all the tests. :-)

test/functional/setup_http_server
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env node
```
erikrose: A grammar nit: "setup" is a noun, "set up" is a verb. Thus, this should be named set_up_http_server. :-)

test/functional/isVisible.js
biancadanforth: I added the eslint rules for this; also I noticed my test files weren't being linted, so I amended `make lint`. Unfortunately, I wasn't able to get the `.mjs` and my `test/functional` `.js` files linted in a single eslint command, so I added a second command.

test/functional/isVisible.js
biancadanforth: Excellent point.

biancadanforth:
Well, that's interesting that this exact same branch would fail here and not in my fathom fork. Anyway, assuming I fix that timing issue some time soon, I also need to do the following to things before removing the "Draft" flag:
* Move the JS functional test work back to its own stage, since coverage doesn't work when executing code via `driver.executeScript`
* Update `npm test` script, so that the functional tests can also be run locally (geckodriver needs to be added to the `$PATH` and the HTTP server needs to start up/shut down)

test/functional/isVisible.html
```diff
@@ -2,86 +2,54 @@
 <html lang="en" dir="ltr">
   <head>
     <meta charset="utf-8">
-    <link rel="shortcut icon" href="">
```
erikrose: What's this <link> for, anyway?

test/functional/isVisible.html
```diff
@@ -2,86 +2,54 @@
...
+    <div id="ancestor-not-visible-7" style="width: 0;">
+      <div id="not-visible-7" style="overflow: hidden;"></div>
+    </div>
+    <!-- ``overflow: hidden`` with zero height ancestor -->
+    <div id="ancestor-not-visible-8" style="height: 0;">
+      <div id="not-visible-8" style="overflow: hidden;"></div>
     </div>
-    <div id="not-visible-8-ancestor">
-      <div id="not-visible-8"></div>
+    <!-- off-screen to the left of viewport -->
```
erikrose: I love how you've commented and inlined these!

test/functional/isVisible.js
```diff
@@ -1,9 +1,9 @@
-const { assert } = require('chai');
+const {assert} = require('chai');
 const firefox = require('selenium-webdriver/firefox');
 const webdriver = require('selenium-webdriver');
```
erikrose: You never actually use ``webdriver`` except to grab more imports out of it. Can you do it all in 1 line?

test/functional/isVisible.js
```diff
@@ -12,45 +12,46 @@ describe('isVisible', () => {
...
+        const isElementVisible = await driver.executeScript(`
+            ${ancestors}
+            ${isDomElement}
+            ${toDomElement}
+            return ${isVisible}(document.getElementById('${id}'));
+        `);
+        assert.equal(
+            isElementVisible,
+            expected,
+            `isVisible should return false for element with id '${id}'.`
```
erikrose: `false` → `${expected}`

test/functional/isVisible.js
```diff
@@ -12,45 +12,46 @@ describe('isVisible', () => {
...
+        );
+    }
+
+    it('should return false when an element is hidden', async () => {
+        const hiddenElementIds = await driver.executeScript(`
```
erikrose: Now that's a nice piece of DRY. :-) The knowledge of how many hidden elements there is lives in only one place: the markup.

test/functional/isVisible.js
```diff
@@ -12,45 +12,46 @@ describe('isVisible', () => {
...
+            return [].slice.call(document.querySelectorAll('[id^="not-visible-"]')).map((element) => element.id);
```
erikrose: Is the `slice` just to copy the results of querySelectorAll into an Array? If so, Array.from(querySelectorAll(...)) might show your intention more clearly.

Shorter still, you could do `Array.prototype.map.call(querySelectorAll(...), ...)`.

test/functional/isVisible.js
```diff
@@ -12,45 +12,46 @@ describe('isVisible', () => {
...
+    it('should return true when an element is visible', async () => {
+        const visibleElementIds = await driver.executeScript(`
+            return [].slice.call(document.querySelectorAll('[id^="visible-"]')).map((element) => element.id);
+        `);
+        await driver.get(TEST_PAGE_URL);
+
+        for (const id of visibleElementIds) {
+            await checkVisibility(id, true);
         }
     }).timeout(60000);
```
erikrose: Btw, I like that you gave isVisible.html and isVisible.js the same root names. That'll make it easier for Future People to untangle that they go together.

test/functional/isVisible.js
```diff
@@ -12,45 +12,46 @@ describe('isVisible', () => {
...
+        }
+    }).timeout(60000);
+
+    it('should return true when an element is visible', async () => {
+        const visibleElementIds = await driver.executeScript(`
+            return [].slice.call(document.querySelectorAll('[id^="visible-"]')).map((element) => element.id);
+        `);
+        await driver.get(TEST_PAGE_URL);
+
+        for (const id of visibleElementIds) {
```
erikrose: Maybe we should even factor all this stuff up (the Array.prototype.... call, the for loop, and everything). It's nontrivial.

test/functional/isVisible.html
biancadanforth: Firefox seems to, by default, make a GET request to get the page's favicon for display in the tab. If I don't have a favicon, my HTTP server throws an error.

This is an empty favicon to prevent that error. I can make a comment saying as much.

Makefile
```diff
@@ -8,6 +8,7 @@ all: $(JS)
 
 lint:
 	@node_modules/.bin/eslint --ext mjs .
+	@node_modules/.bin/eslint test/functional
```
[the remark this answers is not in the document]
biancadanforth: Due to the directory rename, the `isVisible` test now runs first in the `mocha --recursive` command. Not a problem, just an observation.

test/functional/isVisible.js
biancadanforth: I am going to guess what you are saying here, since you have chosen not to use words... heh.

This assertion is not executing inside the page like the `driver.executeScript` call before it. Therefore, I have access to the scope of the `checkVisibility` function, where `expected` is defined.

Also, the first and second arguments into `assert.equal()` are just variables, not template literals.

If I'm missing something, please let me know.

test/functional/isVisible.js
biancadanforth: Hmm... not sure I'm following. Can you clarify?

biancadanforth: Oh I think I see what you are saying -- make a new function that each `it` block calls that gets the ids and checks each element's visibility.

biancadanforth:
@erikrose : This is ready for another look!

I ended up being able to ignore `isVisible` and its called functions with `istanbul`, so I didn't end up moving these functional tests to their own stage. Coverage and the tests are now both working together, which seems preferable. Let me know what you think.

test/functional/isVisible.html
erikrose: gtk! A comment would be great for Future Me.

package.json
```diff
@@ -45,10 +45,13 @@
     "url": "https://github.com/mozilla/fathom.git"
   },
   "scripts": {
+    "add-geckodriver-to-path": "export PATH=$PATH:$PWD/node_modules/geckodriver/bin",
     "coveralls": "cat ./coverage/lcov.info | coveralls",
     "lint": "make lint",
+    "init-server": "./test/functional/start_http_server & echo $! > pid",
     "pretest": "npm run lint",
-    "test": "make coverage"
+    "test": "npm run add-geckodriver-to-path && npm run init-server && make coverage && npm run uninit-server",
```
erikrose: You can actually keep the shell environment cleaner by avoiding the `export`: if you say this instead, the shell variable change will be in scope only for this one line:

```
PATH=$PATH:$PWD/node_modules/geckodriver/bin npm run add-geckodriver-to-path && npm run init-server && make coverage && npm run uninit-server
```

The only trouble is that none of this shell cleverness will work under Windows, where @danielhertenstein lives. I can't really think of a way to make it work, either, without a lot of extra code. Let's leave it like this for now and see if it causes him trouble. After all, `make test` should still run all but the browser tests, and he does have the option of falling back to a WSL shell if he wants to run those too.

test/functional/isVisible.js
biancadanforth: Oh; yes the error message if the check fails. Thanks for the catch.

package.json
biancadanforth: > You can actually keep the shell environment cleaner by avoiding the `export`: if you say this instead, the shell variable change will be in scope only for this one line:
> ```
> PATH=$PATH:$PWD/node_modules/geckodriver/bin npm run add-geckodriver-to-path && npm run init-server && make coverage && npm run uninit-server
> ```

Did you mean to remove `npm run add-geckodriver-to-path`? This is what I have now that is working:

```diff
diff --git a/package.json b/package.json
index bb87939..fa2a73c 100644
--- a/package.json
+++ b/package.json
@@ -45,12 +45,11 @@
     "url": "https://github.com/mozilla/fathom.git"
   },
   "scripts": {
-    "add-geckodriver-to-path": "export PATH=$PATH:$PWD/node_modules/geckodriver/bin",
     "coveralls": "cat ./coverage/lcov.info | coveralls",
     "lint": "make lint",
     "init-server": "./test/browser/start_http_server & echo $! > pid",
     "pretest": "npm run lint",
-    "test": "npm run add-geckodriver-to-path && npm run init-server && make coverage && npm run uninit-server",
+    "test": "PATH=$PATH:$PWD/node_modules/geckodriver/bin && npm run init-server && make coverage && npm run uninit-server",
     "uninit-server": "kill $(cat pid) && rm pid"
   },
   "main": "index"
```

> The only trouble is that none of this shell cleverness will work under Windows, where @danielhertenstein lives. I can't really think of a way to make it work, either, without a lot of extra code. Let's leave it like this for now and see if it causes him trouble. After all, `make test` should still run all but the browser tests, and he does have the option of falling back to a WSL shell if he wants to run those too.

Ah that's a good point. Actually, I had previously added the `--recursive` option to the `make test` command, which means it would try to run the JS functional tests at `test/browser` as well. It breaks currently, since I'm not doing the requisite set-up and clean-up. I'll remove that option for now, so `make test` doesn't run the JS functional tests at all. I'll add a comment to note this distinction in the Makefile.

erikrose:
Lgtm!

erikrose:
Please mention "Fix #122" in the merge commit when you merge this.

PR #148, opened 2019-08-30
Fix #138: Add fathom-label CLI

### LINKED ISSUE #138: Create a CLI tool to automate tagging an entire page
Some applications of Fathom are interested in categorizing an entire page (e.g. shopping pages or article pages). Since page tagging is tedious and error-prone, it would be helpful to have an automated tool to help in this situation, where there is no ambiguity as to which element to tag.

We have discussed tagging the `<html>` element for this purpose. Can we make a `fathom-tag` or similar CLI that takes in the desired tag value (e.g. `"article"`) and adds a `data-fathom` attribute with that value to each page's `<html>` element?

biancadanforth:
* This is largely based off of the 'fathom-extract' CLI.
* This tool handles documents with multiple 'html' elements by capturing each original opening html tag and each modified opening html tag in a list of tuples and iterating through each tuple to 'string.replace' to update the HTML document.
* I originally had a test that asserted that '< html >' became '< html data-fathom="test" >', but I discovered that this isn't valid HTML per the [specification](https://www.w3.org/TR/html53/syntax.html#start-tags)
* Similarly, there is no test for comments inside an 'html' tag, because that is also invalid HTML.

Open question: Is `fathom-label` descriptive enough? This tool only labels an entire page, so perhaps we can be more specific. Suggestions welcome!

### DIFF
--- a/cli/fathom_web/commands/label.py
+++ b/cli/fathom_web/commands/label.py
@@ -1,0 +1,87 @@
+from html.parser import HTMLParser
+import pathlib
+import re
+import shutil
+
+from click import argument, command, option, Path, STRING
+
+
+@command()
+@option('--preserve-originals/--no-preserve-originals',
+        default=True,
+        help='Save original HTML files in a newly created `originals`'
+             ' directory in IN_DIRECTORY (default: True)')
+@argument('in_directory', type=Path(exists=True, file_okay=False))
+@argument('in_type', type=STRING)
+def main(in_directory, in_type, preserve_originals):
+    """
+    Add the ``data-fathom`` attribute with a value of IN_TYPE to the
+    opening tag of any ``<html>`` elements in the HTML pages in
+    IN_DIRECTORY. This tool is used to label an entire webpage (e.g.
+    IN_TYPE could be "article" for article webpages).
+    """
+    if preserve_originals:
+        originals_dir = pathlib.Path(in_directory) / 'originals'
+        try:
+            originals_dir.mkdir(parents=True)
+        except FileExistsError:
+            raise RuntimeError(f'Tried to make directory {originals_dir.as_posix()}, but it already exists. To protect'
+                               f' against unwanted data loss, please move or remove the existing directory.')
+    else:
+        originals_dir = None
+
+    for file in pathlib.Path(in_directory).iterdir():
+        if file == originals_dir:
+            continue
+        if file.is_dir():
+            print(f'Skipping directory {file.name}/')
+            continue
+        if file.suffix != '.html':
+            print(f'Skipping {file.name}; not an HTML file')
+            continue
+
+        with file.open(encoding='utf-8') as fp:
+            html = fp.read()
+
+        new_html = label_html_tags_in_html_string(html, in_type)
+
+        if preserve_originals:
+            shutil.move(file, originals_dir / file.name)
+
+        with file.open('w', encoding='utf-8') as fp:
+            fp.write(new_html)
+
+
+class HTMLParserSubclass(HTMLParser):
+    def __init__(self, in_type, **kwargs):
+        self.in_type = in_type
+        self.html_tags_list = []
+        super().__init__(**kwargs)
+
+    def handle_starttag(self, tag, attrs):
+        if tag == 'html':
+            original_html_tag = self.get_starttag_text()
+            new_html_substring = f'html data-fathom="{self.in_type}"'
+            new_html_tag = original_html_tag.replace('html', new_html_substring, 1)
+            self.html_tags_list.append(tuple([original_html_tag, new_html_tag]));
+
+
+def label_html_tags_in_html_string(html: str, in_type: str) -> str:
+    """
+    Finds all opening ``html`` tags in the HTML string and adds a
+    ``' data-fathom="${in_type}"'`` substring to each one.
+
+    We do this by building a new HTML string with the inserted substring(s).
+
+    The ``html`` tags are found using the HTMLParser class in Python's
+    built-in html.parser library.
+    """
+    parser = HTMLParserSubclass(in_type)
+    parser.feed(html);
+
+    new_html = html
+
+    for (original_html_tag, new_html_tag) in parser.html_tags_list:
+        new_html = new_html.replace(original_html_tag, new_html_tag, 1);
+
+    return new_html
--- a/cli/fathom_web/commands/label.py
+++ b/cli/fathom_web/commands/label.py
@@ -1,9 +1,10 @@
 from html.parser import HTMLParser
+import os
 import pathlib
 import re
 import shutil
 
-from click import argument, command, option, Path, STRING
+from click import argument, command, option, Path, STRING, progressbar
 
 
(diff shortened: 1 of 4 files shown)


biancadanforth:
@erikrose Ready for a look now. I was able to get multiprocessing to work, but I did it with the `multiprocessing` library, as I had an easier time of it since I'd used it before with mythmon than `concurrent.futures`. With multiprocessing, I see a 72% faster running time for 200 HTML pages.

cli/fathom_web/commands/label.py
```diff
@@ -0,0 +1,112 @@
...
+    new_html = label_html_tags_in_html_string(html, in_type)
+
+    if preserve_originals:
+        shutil.move(file, originals_dir / file.name)
+
+    with file.open('w', encoding='utf-8') as fp:
+        fp.write(new_html)
+
+
+class HTMLParserSubclass(HTMLParser):
```
danielhertenstein: This is personal preference, but I bring it up in case you like style too: Consider moving this class below `label_html_tags_in_html_string()` (where it is used). When reading code for the first time, I find it helpful to see where/how something is used before its definition. Otherwise, I think, "Why does this exist?"

erikrose: I admit I am inconsistent about this. I always put main() at the bottom because I grew up with Pascal, but I'm open to changing that.

cli/fathom_web/commands/label.py
```diff
@@ -0,0 +1,112 @@
...
+        originals_dir = None
+
+    list_of_items = os.listdir(in_directory)
+    number_of_items = len(list_of_items)
+
+    print_statements = [] # Capture any print statements to log at the end.
+
+    # Make a pool of workers. Each worker is in its own process. We use a scaling factor to account for the overhead of
+    # setting up all of the processes.
+    pool = multiprocessing.Pool(multiprocessing.cpu_count() * 4)
```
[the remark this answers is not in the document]
erikrose: Where did the 4 come from? My CPUs, at least, have 2 logical cores each.

cli/fathom_web/commands/label.py
```diff
@@ -0,0 +1,112 @@
...
+        self.in_type = in_type
+        self.html_tags_list = []
+        super().__init__(**kwargs)
+
+    def handle_starttag(self, tag, attrs):
+        if tag == 'html':
+            original_html_tag = self.get_starttag_text()
+            new_html_substring = f'html data-fathom="{self.in_type}"'
+            new_html_tag = original_html_tag.replace('html', new_html_substring, 1)
+            self.html_tags_list.append(tuple([original_html_tag, new_html_tag]))
```
erikrose: It's rarely necessary to call `tuple()`; you can spell a literal tuple like this: `(original_html_tag, new_html_tag)`.

cli/fathom_web/commands/label.py
```diff
@@ -0,0 +1,112 @@
...
+    The ``html`` tags are found using the HTMLParser class in Python's
+    built-in html.parser library.
+    """
+    parser = HTMLParserSubclass(in_type)
+    parser.feed(html)
+
+    new_html = html
+
+    for (original_html_tag, new_html_tag) in parser.html_tags_list:
+        new_html = new_html.replace(original_html_tag, new_html_tag, 1)
```
erikrose: This actually makes us do the work of scanning through the string twice: once in the parser, and again here, to find the previously found string and replace it. Did my suggestion of HTMLParser.getpos() not work out?

cli/setup.py
```diff
@@ -18,6 +18,7 @@
         'fathom-list = fathom_web.commands.list:main',
         'fathom-pick = fathom_web.commands.pick:main',
         'fathom-serve = fathom_web.commands.serve:main',
+        'fathom-label = fathom_web.commands.label:main',
```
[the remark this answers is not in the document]
biancadanforth: Yeah I'd originally called the tool `fathom-tag`.

cli/fathom_web/commands/label.py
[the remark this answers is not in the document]
biancadanforth: `cpu_count()` returns the number of logical cores, so the `4` is not the ratio of logical cores to physical cores. It is an overhead factor that mythmon suggested, since there is also work associated with setting up the IPC and for task switching.

I can make the number of workers an option; that's a good idea.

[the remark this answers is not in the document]
biancadanforth: I can actually remove the `4`, since my husband plotted the speed of the script as a function of number of processes, and beyond 8 or 9 showed little to no gains... Well, at least my hardware has 8 logical cores...

cli/fathom_web/commands/label.py
```diff
@@ -0,0 +1,112 @@
...
+
+    print_statements = [] # Capture any print statements to log at the end.
+
+    # Make a pool of workers. Each worker is in its own process. We use a scaling factor to account for the overhead of
+    # setting up all of the processes.
+    pool = multiprocessing.Pool(multiprocessing.cpu_count() * 4)
+    # Curry ``task``, so we can pass more than one argument into pool.imap_unordered.
+    task = partial(label_task, in_directory, in_type, originals_dir, preserve_originals)
+
+    with progressbar(pool.imap_unordered(task, list_of_items),
```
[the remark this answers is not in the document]
biancadanforth: Hmm good point, I forgot about that optional argument of `chunksize`.

I experimented with this a bit just now for the 200 HTML pages, and it looks like 5 could be marginally faster, though that difference could also be random variation.

| `chunksize`  | average time (s) for 3 runs |
| ------------- | ------------- |
| 1  | 14.635  |
| 5  | 14.346  |
| 10  | 15.097  |
| 50  | 19.781  |

cli/fathom_web/commands/label.py
biancadanforth: Hm... yeah. I've learned to arrange these the exact opposite way in JavaScript (though looking here I am pretty inconsistent too :P): that a declaration should occur before it is used. I don't have a strong preference here, so I can move it.

cli/fathom_web/commands/label.py
biancadanforth: I couldn't figure out how to do it with `getpos()`. Inside `handle_starttag`, if I call `self.getpos()`, it returns `(2, 0)` -- i.e. line 2, with offset of 0 -- i.e. the line number and offset for the opening `<html>` tag.

Maybe I am misunderstanding how to use [`getpos()`](https://docs.python.org/3/library/html.parser.html#html.parser.HTMLParser.getpos), but I don't see how it can be used to give me the offset within the opening `<html>` tag for where to insert the new `data-fathom` attribute.

biancadanforth: Even if I could get that, I don't see a method in the class to replace the opening `<html>` tag with a new value.

erikrose: Alright, that means this wasn't the best choice of parser for this job. But it's good enough for research code; it's not like we're running this in tight loops.

PR #150, opened 2019-09-05
Add Python linter. Fix #149.

### LINKED ISSUE #149: Add linting for at least Python
This saves review roundtrips for trivia like accidental semicolons.

erikrose:
And rejigger the build process in the…well…process.

If this all works out, here's what changes from a dev point of view:
* The makefile is now the One True Entrypoint for build stuff. There are no more npm scripts.
* `make test` now runs all the tests, even the browser ones.
* The browser tests now work on Windows.
* `make lint` lints all languages. `make py_lint` and `make js_lint` lint only 1.
* `make py_test` and `make js_test` test 1 language each.
* `make` takes care of making a venv for you (in the top level of the checkout) whenever it needs one. If you have an existing one activated before you make, it'll use yours instead.
* New Python dependencies are automatically installed at the next `make` whenever they're added to requirements files or setup.py.
* `npm install` is run automatically whenever package.json has been updated.
* `make doc` from the top level now builds the docs.
* A failing doc build will now fail the CI tests, so we don't get surprised on master anymore.

Note that you'll see occasional spurious package installation attempts after you change branches, because the branch change causes the mod dates of files to be reset to the current time. But the attempts are reasonably quick and idempotent.


Makefile
```diff
@@ -1,3 +1,7 @@
+# If there's an activated virtualenv, use that. Otherwise, make one in the cwd.
+VIRTUAL_ENV ?= $(PWD)/venv
+PATH := $(PWD)/node_modules/geckodriver/bin:$(PATH)
```
erikrose: I'll be darned if I can get this scoped tighter. Only all_js_test needs it.

Makefile
```diff
@@ -28,7 +43,30 @@ publish: $(JS)
...
+
+
+# Private targets:
+
+# Make a virtualenv at $VIRTUAL_ENV if there isn't one or if requirements have
+# changed. Install the dev requirements and the actual requirements.
+$(VIRTUAL_ENV)/pyvenv.cfg: tooling/dev-requirements.txt cli/setup.py
+	$(PYTHON3) -m venv $(VIRTUAL_ENV)
+	pip3 install -r tooling/dev-requirements.txt
+	pip3 install -e cli
```
danielhertenstein: Windows + pip requires: `pip3 install -e cli -f https://download.pytorch.org/whl/torch_stable.html` The addition of `dependency_links` only covers the `python setup.py install` case.

Makefile
```diff
@@ -1,23 +1,38 @@
...
-	@node_modules/.bin/mocha
-	pytest cli/fathom_web/test
+py_lint: $(VIRTUAL_ENV)/pyvenv.cfg
+	flake8 cli
+
+test: js_test py_test
 
-coverage: $(JS)
+js_test: $(JS)
 	@node_modules/.bin/istanbul cover node_modules/.bin/_mocha -- --recursive
```
danielhertenstein: Windows requires: `@node_modules/.bin/istanbul cover node_modules/mocha/bin/_mocha -- --recursive` as noted [here](https://github.com/gotwarlost/istanbul#usage-on-windows). This was a problem I ran into when I first started and had as a local change and completely forgot about.

erikrose: Huh. Then I'm surprised .bin/eslint and .bin/babel work. Do they? They're all symlinks on my box.

Makefile
erikrose: I'm afraid that really is the case. :-( They seem to have removed both the `--process-dependency-links` and the `--allow-all-external` options, making this repetition necessary.

Makefile
erikrose: Ah, reading the link made the reasons clear.

Makefile
erikrose: What's more, https://www.python.org/dev/peps/pep-0508/, which is the closest thing to a replacement, doesn't support the flexibilty we need in its URL-based requirements: it provides only the ability to link directly to a specific file. We don't want to pin down the dependency to that degree. So, old cruft it is!

PR #151, opened 2019-09-06
Remove the Readability tests, which weren't paying their rent. Finally remove the old optimizer, which was used only by those tests.

erikrose:
* Tests are now 35s faster.
* Code coverage actually goes *up* for all counts; it wasn't really testing anything but itself:

  Before:
  Statements   : 82.53% ( 737/893 )
  Branches     : 73.02% ( 230/315 )
  Functions    : 86.04% ( 228/265 )
  Lines        : 82.24% ( 699/850 )

  After:
  Statements   : 86.17% ( 648/752 )
  Branches     : 71.68% ( 200/279 )
  Functions    : 84.58% ( 192/227 )
  Lines        : 86.97% ( 621/714 )
* We get to throw away a whole bunch of test corpus.
* We can remove dirsIn(), too, which is now dead.


erikrose:
Rebased and pushed as c3fec39c90be2a496bace0f1e9d596c74dd51fbb.

PR #152, opened 2019-09-06
Updates decode to work on unpadded base64 strings

danielhertenstein:
I encountered frozen pages with unpadded base64 strings. Python's `base64.b64decode()` method requires padded strings, so this PR pads strings that are unpadded.

@biancadanforth: You may need this for your samples.

### DIFF
--- a/cli/fathom_web/commands/extract.py
+++ b/cli/fathom_web/commands/extract.py
@@ -163,10 +163,20 @@ def generate_filename(mime_type: str, filename: str) -> str:
 
 def decode(base64_string: str) -> bytes:
     """
-    Decodes the base64 string into bytes. If as string has any additional
-    encoding (ex. percent encoding of the padding characters), decode that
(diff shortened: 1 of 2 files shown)


cli/fathom_web/commands/extract.py
```diff
@@ -163,10 +163,20 @@ def generate_filename(mime_type: str, filename: str) -> str:
...
+    We also check if the string is padded to a number of characters that is a
+    multiple of four, which is what base64.b64decode() requires.
     """
+    # Percent encoding
     if '%' in base64_string:
         base64_string = unquote(base64_string)
+    # Padding check
+    string_mod_4 = len(base64_string) % 4
+    if string_mod_4 != 0:
+        base64_string += "=" * (4 - string_mod_4)
```
erikrose: single quotes :-). The linter should soon point this out.

PR #154, opened 2019-09-25
Specializes fathom-list to create files more useful for FathomFox

danielhertenstein:
This PR makes `fathom-list` a specialized tool for FathomFox. I have found that I have needed to do a significant amount of work to create the filename lists I needed for vectorization, so I'm proposing `fathom-list` do all this work for us.

It is likely that training samples are not all in the same directory. You may the following directory structure (which is what we're currently recommending):
```
samples/
|-  training/
|    |-  source_a/
|    |    |-  positive/
|    |    |    |-  [lots of html files]
|    |    |-  negative/
|    |    |    |-  [lots of html files]
|    |-  source_b/
|    |    |-  positive/
|    |    |    |-  [lots of html files]
|    |    |-  negative/
|    |    |    |-  [lots of html files]
```
In this case, using `fathom-list` is not convenient for getting your sample filenames since the FathomFox wants paths relative to some base directory. In the past, I've served `samples/` as my base directory using `fathom-serve` so don't have to change it when switching between training and validation sets. Since I'm serving `samples/` my filenames need to be paths relative to `samples/`. To get this list I do the following:

1. Use `fathom-list` to list the files in one of the directories containing HTML files and save it to a file
2. Remove the `resources/` line from the file created in 1 since `fathom-list` lists all files.
3. Use `fathom-list` to list the files in each of the other directories and manually paste the filenames into the file I created in 1.
4. Prepend each filename with the path relative to `samples/`. I at least use a vim macro to do this part, though the macro has to change for each directory.

This proposed new `fathom-list` does this for you automatically. You can specify the base directory to make all the outputted paths relative to. You can also do the listing recursively. Now to generate the filenames for my training set, I run the following from my `samples/` directory:
```
fathom-list training -b . -r -o training.txt
```
This lists paths relative to `samples/` for all the HTML files in `training/`, recursively, and saves them to the file `training.txt`.

If this seems like a positive change, I'll write up the tests for this and submit a full PR

### DIFF
--- a/cli/fathom_web/commands/list.py
+++ b/cli/fathom_web/commands/list.py
@@ -4,30 +4,40 @@ from click import argument, command, File, option, Path
 
 
 @command()
+@argument('in_directory', type=Path(exists=True, file_okay=False))
+@option('--base-dir', '-b', type=Path(exists=True, file_okay=False),
+        help='The directory to create relative paths from.')
+@option('--recursive', '-r', default=False, is_flag=True,
+        help='Recursively list files from the IN_DIRECTORY and all subdirectories.')
 @option('--out-file', '-o', type=File(mode='w'), default=None,
         help='A file for saving the printed filenames for easy future reference.')
-@argument('in_directory', type=Path(exists=True, file_okay=False))
-def main(in_directory, out_file):
+def main(in_directory, base_dir, recursive, out_file):
     """
-    Lists filenames in a IN_DIRECTORY, one filename per line. Optionally saves output to OUT_FILE.
-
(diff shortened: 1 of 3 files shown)


biancadanforth:
I tried this out for the Smoot Articles samples, and it worked great, since my file structure mirrors the file structure you used for your Shopping samples. Seems like a good idea!

@danielhertenstein let me know when the PR is ready for review, since it's marked as a Draft and you had mentioned you wanted to add tests.

danielhertenstein:
Hooray! I'll be adding the tests when I get back from vacation.

PR #156, opened 2019-10-23
Make `fathom list` a specialized tool for FathomFox

danielhertenstein:
This PR makes `fathom-list` a specialized tool for FathomFox. I have found that I have needed to do a significant amount of work to create the filename lists I needed for vectorization, so I'm proposing `fathom-list` do all this work for us.

It is likely that training samples are not all in the same directory. You may the following directory structure (which is what we're currently recommending):
```
samples/
|-  training/
|    |-  source_a/
|    |    |-  positive/
|    |    |    |-  [lots of html files]
|    |    |-  negative/
|    |    |    |-  [lots of html files]
|    |-  source_b/
|    |    |-  positive/
|    |    |    |-  [lots of html files]
|    |    |-  negative/
|    |    |    |-  [lots of html files]
```
In this case, using `fathom-list` is not convenient for getting your sample filenames since the FathomFox wants paths relative to some base directory. In the past, I've served `samples/` as my base directory using `fathom-serve` so don't have to change it when switching between training and validation sets. Since I'm serving `samples/` my filenames need to be paths relative to `samples/`. To get this list I do the following:

1. Use `fathom-list` to list the files in one of the directories containing HTML files and save it to a file
2. Remove the `resources/` line from the file created in 1 since `fathom-list` lists all files.
3. Use `fathom-list` to list the files in each of the other directories and manually paste the filenames into the file I created in 1.
4. Prepend each filename with the path relative to `samples/`. I at least use a vim macro to do this part, though the macro has to change for each directory.

This proposed new `fathom-list` does this for you automatically. You can specify the base directory to make all the outputted paths relative to. You can also do the listing recursively. Now to generate the filenames for my training set, I run the following from my `samples/` directory:
```
fathom-list training -b . -r -o training.txt
```
This lists paths relative to `samples/` for all the HTML files in `training/`, recursively, and saves them to the file `training.txt`.

### DIFF
--- a/cli/fathom_web/commands/list.py
+++ b/cli/fathom_web/commands/list.py
@@ -41,7 +41,6 @@ def main(in_directory, base_dir, recursive, out_file):
         if out_file is not None:
             filenames_to_save.append(relative_path.as_posix() + '\n')
 
-
     if out_file is not None:
         if there_were_no_files:
             print(f'No .html files found in {in_directory}. Did not create {out_file}.')
--- a/cli/fathom_web/commands/list.py
+++ b/cli/fathom_web/commands/list.py
@@ -32,12 +32,18 @@ def main(in_directory, base_dir, recursive, out_file):
     if out_file is not None:
         filenames_to_save = []
 
+    there_were_no_files = True
     for file in path_iterator:
+        there_were_no_files = False
         relative_path = file.relative_to(base_dir)
         print(relative_path)
 
         if out_file is not None:
             filenames_to_save.append(relative_path.as_posix() + '\n')
 
+
     if out_file is not None:
-        out_file.writelines(filenames_to_save)
+        if there_were_no_files:
+            print(f'No .html files found in {in_directory}. Did not create {out_file}.')
+        else:
+            out_file.writelines(filenames_to_save)
--- a/cli/fathom_web/commands/list.py
+++ b/cli/fathom_web/commands/list.py
@@ -4,30 +4,40 @@ from click import argument, command, File, option, Path
 
 
 @command()
+@argument('in_directory', type=Path(exists=True, file_okay=False))
+@option('--base-dir', '-b', type=Path(exists=True, file_okay=False),
+        help='The directory to create relative paths from.')
+@option('--recursive', '-r', default=False, is_flag=True,
+        help='Recursively list files from the IN_DIRECTORY and all subdirectories.')
 @option('--out-file', '-o', type=File(mode='w'), default=None,
         help='A file for saving the printed filenames for easy future reference.')
-@argument('in_directory', type=Path(exists=True, file_okay=False))
-def main(in_directory, out_file):
+def main(in_directory, base_dir, recursive, out_file):
     """
-    Lists filenames in a IN_DIRECTORY, one filename per line. Optionally saves output to OUT_FILE.
-
-    Ignores hidden files.
+    Lists relative paths of HTML files in a IN_DIRECTORY relative to BASE_DIR, one filename per line.
+    If BASE_DIR is not specified, paths are relative to IN_DIRECTORY. Optionally saves output to OUT_FILE.
+    Optionally performs the listing recursively.
 
     This is useful for vectorizing samples using FathomFox. FathomFox expects input filenames copied into a text box
-    with one filename per line.
+    with one filename per line and relative to some path you are serving files from using fathom-serve.
     """
+    if base_dir is None:
+        base_dir = in_directory
+    base_dir = pathlib.Path(base_dir)
+
+    if recursive:
+        path_iterator = pathlib.Path(in_directory).rglob('*.html')
+    else:
+        path_iterator = pathlib.Path(in_directory).glob('*.html')
+
     if out_file is not None:
         filenames_to_save = []
 
-    for file in pathlib.Path(in_directory).iterdir():
-        # Ignore hidden files
-        if file.name.startswith('.'):
-            continue
-
-        print(file.name)
+    for file in path_iterator:
+        relative_path = file.relative_to(base_dir)
+        print(relative_path)
 
         if out_file is not None:
-            filenames_to_save.append(file.name + '\n')
+            filenames_to_save.append(relative_path.as_posix() + '\n')
 
     if out_file is not None:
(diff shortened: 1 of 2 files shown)


cli/fathom_web/test/test_list.py
```diff
@@ -0,0 +1,176 @@
...
+        [
+            in_directory.as_posix(),
+            f'-o{out_file.as_posix()}',
+        ],
+    )
+    assert result.exit_code == 0
+
+    expected_file_contents = ""
+    actual_file_contents = out_file.read_text()
+    assert expected_file_contents == actual_file_contents
```
biancadanforth: If there are no .html files in `in_directory`, do we want to create a .txt file at all? Maybe we could just print a statement like: "No .html files found in ${in_directory}. Did not create ${out_file}." Then you could assert the correct, non-zero exit code.

cli/fathom_web/test/test_list.py
```diff
@@ -0,0 +1,176 @@
...
+    # Run fathom-list
+    result = CliRunner().invoke(
+        list_main,
+        [
+            in_directory.as_posix(),
+            f'-b{base_dir.as_posix()}',
+            '-r',
+            f'-o{out_file.as_posix()}',
+        ],
+    )
```
biancadanforth: This is cool! TIL how to simulate a command for a CLI test.

cli/fathom_web/test/test_list.py
```diff
@@ -0,0 +1,176 @@
...
+            in_directory.as_posix(),
+            f'-b{base_dir.as_posix()}',
+            f'-o{out_file.as_posix()}',
+        ],
+    )
+    assert result.exit_code == 0
+
+    expected_file_contents = ""
+    actual_file_contents = out_file.read_text()
+    assert expected_file_contents == actual_file_contents
```
biancadanforth: Same here for throwing an error instead of creating an empty file. Or if you don't throw an error, perhaps post a warning before creating the empty output file?

cli/fathom_web/test/test_list.py
```diff
@@ -0,0 +1,176 @@
...
+    assert result.exit_code == 0
+
+    expected_file_contents = [
+        a1.relative_to(base_dir).as_posix(),
+        a2.relative_to(base_dir).as_posix(),
+        b1.relative_to(base_dir).as_posix(),
+        b2.relative_to(base_dir).as_posix(),
+    ]
+    actual_file_contents = out_file.read_text().splitlines()
+    assert expected_file_contents == actual_file_contents
```
biancadanforth: This test is failing for me (see [verbose pytest output](https://paste.mozilla.org/NLTQUNU0)).

```
cli/fathom_web/test/test_list.py:39: AssertionError
------------------------------- Captured stdout call --------------------------------
actual_file_contents: $['in_directory/source_a/2.html', 'in_directory/source_a/1.html', 'in_directory/source_b/2.html', 'in_directory/source_b/1.html']
expected_file_contents: $['in_directory/source_a/1.html', 'in_directory/source_a/2.html', 'in_directory/source_b/1.html', 'in_directory/source_b/2.html']
```

It looks like checking two lists for equality in Python requires the list elements to have the same values in the same order, but in my case, the values are the same but the order is different -- though I'm not sure why this doesn't happen for you or for the Travis machine. Since order doesn't matter in this case, can we check for the same values in any order instead?
I have to imagine perhaps the way my OS does recursive search could be different?

cli/fathom_web/test/test_list.py
```diff
@@ -0,0 +1,176 @@
...
+    assert result.exit_code == 0
+
+    expected_file_contents = [
+        a1.relative_to(in_directory).as_posix(),
+        a2.relative_to(in_directory).as_posix(),
+        b1.relative_to(in_directory).as_posix(),
+        b2.relative_to(in_directory).as_posix(),
+    ]
+    actual_file_contents = out_file.read_text().splitlines()
+    assert expected_file_contents == actual_file_contents
```
biancadanforth: This assertion is also failing for me, because my `actual_file_contents` are the same values in a different order.

cli/fathom_web/test/test_list.py
danielhertenstein: It took me a crazy long time to get this right. The first time I went through this I had to trace through the execution of click to find that you need to pass strings for the arguments. This time, I found out an extremely hard way that you cannot have spaces in the strings. So `f'-b{base_dir.as_posix()}'` cannot be `f'-b {base_dir.as_posix()}'`. click will not strip the whitespace, so you get an OS error about an invalid argument to `open()`, but when the error is reported, I didn't notice the extra whitespace in the message.

cli/fathom_web/test/test_list.py
danielhertenstein: Very interesting. I'll switch it to a set since that's conceptual more appropriate.

cli/fathom_web/test/test_list.py
danielhertenstein: That's a great idea. I'll make that change.

cli/fathom_web/commands/list.py
```diff
@@ -32,12 +32,18 @@ def main(in_directory, base_dir, recursive, out_file):
...
         print(relative_path)
 
         if out_file is not None:
             filenames_to_save.append(relative_path.as_posix() + '\n')
 
+
     if out_file is not None:
-        out_file.writelines(filenames_to_save)
+        if there_were_no_files:
+            print(f'No .html files found in {in_directory}. Did not create {out_file}.')
```
biancadanforth: After trying this out, it looks like `out_file` here is not a string, but a [`click.File`](https://click.palletsprojects.com/en/7.x/api/#click.File), so when I trigger this if block, I get: `No .html files found in dummy. Did not create <unopened file 'training.txt' w>.` Is it possible just to get the filename string? Otherwise we could just say "Did not create the output file."

danielhertenstein:
What does R+WC mean?

biancadanforth:
> What does R+WC mean?
That's an R+ with changes. I got that lingo from Osmose, your predecessor, so you can thank him. :-)

Basically, the change I am requesting is minor enough that I think it's okay for you to make it and merge this PR without requiring another review cycle.

PR #157, opened 2019-10-29
Adds confidence intervals for false positives and false negatives

danielhertenstein:
As the title suggests, this PR adds 95% confidence intervals to the output of the false positive and false negative metrics computed by fathom-train.

### DIFF
--- a/cli/fathom_web/accuracy.py
+++ b/cli/fathom_web/accuracy.py
@@ -143,9 +143,14 @@ def accuracy_per_page(model, pages):
     return (successes / len(pages)), '\n'.join(report_lines)
 
 
-def pretty_accuracy(description, accuracy, number_of_samples, false_positive=None, false_negative=None):
+def pretty_accuracy(description, accuracy, number_of_samples, false_positive=None, false_negative=None, number_of_positives=None):
     ci_low, ci_high = confidence_interval(accuracy, number_of_samples)
-    falses = '' if false_positive is None else '  FP: {false_positive:.3f}  FN: {false_negative:.3f}'.format(false_positive=false_positive, false_negative=false_negative)
+    if false_positive is not None:
+        fp_ci_low, fp_ci_high = confidence_interval(false_positive, number_of_positives)
+        fn_ci_low, fn_ci_high = confidence_interval(false_negative, number_of_samples - number_of_positives)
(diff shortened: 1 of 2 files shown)


erikrose:
Hope you don't mind me going ahead and merging this; I wanted to base some more changes on it.

biancadanforth:
Using this as an opportunity to get a refresher on statistics! So if I have something output now like this:
```
Training accuracy per tag:  0.92500    95% CI: (0.84337, 1.00000)
                         FP:  0.02500    95% CI: (0.00000, 0.09342)
                         FN:  0.05000    95% CI: (0.00000, 0.14552)
```
This is saying, for example, that there is a 95% chance that the true false positive rate for _all_ pages is between 0.00000 and 0.09342 (or 0% and 9.3%), assuming that our samples are randomly selected, we have a large enough sample size, etc. Is that right?

erikrose:
> that there is a 95% chance that the true false positive rate for all pages is between 0.00000 and 0.09342 (or 0% and 9.3%), assuming that our samples are randomly selected,

Yes, except of course it's per tag in this instance. So, it means that we're 95% sure that the real ratio of false-positive tag categorizations is in (0.00000, 0.09342).

> What does a "good" 95% CI look like for:
>    Training accuracy per tag
>    FP
>    FN

Well, since "good" is a subjective term, I'll give my personal preference. :-) I like training accuracy per tag to be (≥.9, …). In the case you pasted, I'd probably go out and add more samples until I could hit that.

I don't consider FP and FN to be as important, but, if they were wildly misbalanced, like one being 10x the other, I'd want to stop and see if I could explain why that was so. And, of course, in some applications, it's vital to have your error more on one side that another. We can theoretically balance that by changing a (hard-coded at the moment) number in the trainer (see `pos_weight` in the PyTorch docs). It didn't do what it said last time I tried it, though.

biancadanforth:
Okay thanks for clarifying. 🤔 Cool. I'm working on bringing the lower bound of the 95% confidence interval on training accuracy per tag to be 0.9. We shall see!

erikrose:
Note that we currently use the [normal approximation method](https://en.wikipedia.org/wiki/Binomial_proportion_confidence_interval#Normal_approximation_interval) to compute the CI and that it "is unreliable when the sample size is small or the success probability is close to 0 or 1". We should probably switch to the Wilson score interval when we get a chance.

biancadanforth:
I know small is relative; in past statistics classes I have heard n > 40, but what is small? And for success probability, what is meant by that?

That seems worth filing an issue for if so.

erikrose:
As for what is small, you know as much as I!

Success probability merely refers to confidence, in our context. The CI math is all modeled on "successes" or "failures" on boolean trials, which we equate with "member of the class" or "not".

PR #161, opened 2019-11-07
HTTPS server becomes HTTP server

danielhertenstein:
Addresses mozilla/fathom-fox#56. We don't actually need the S portion of the HTTPS server, so this PR removes it and the associate cert file


cli/fathom_web/commands/serve.py
```diff
@@ -14,22 +12,14 @@
...
     """
     with cd(directory):
         server = HTTPServer(('localhost', port), SimpleHTTPRequestHandler)
-        certfile = resource_filename('fathom_web', 'cert.pem')
-        server.socket = ssl.wrap_socket(
-            server.socket,
-            certfile=certfile,
-            server_side=True,
-        )
         print(f'Serving {directory} over https://localhost:{port}.')
```
erikrose: https→http

cli/fathom_web/commands/serve.py
```diff
@@ -14,22 +12,14 @@
         help='The directory to serve files from (default: current working directory')
 def main(directory, port):
     """
-    Serves the files in DIRECTORY over a local HTTPS server: https://localhost:PORT.
+    Serves the files in DIRECTORY over a local HTTP server: https://localhost:PORT.
```
erikrose: https→http

danielhertenstein:
Bah, thanks for catching the ones I missed.

PR #164, opened 2019-11-18
`fathom-vectorize` experimental version

danielhertenstein:
This PR adds an experimental version of `fathom-vectorize`. The tools allows a user to create feature vectors from a set of samples using FathomFox's vectorizer, but from the comfort of the command line and with a single command. I say this is an experimental version because more and better error handling is needed. There are TODOs in the code, and I will add issues as necessary following approval of this PR but before merging.

This PR also removes testing for Python 3.6 because there are some features used in `fathom-vectorize` that were added to the language in Python 3.7.


cli/fathom_web/commands/vectorize.py
```diff
@@ -0,0 +1,315 @@
...
+import time
+import zipfile
+
+from click import argument, command, option, Path, progressbar
+from selenium import webdriver
+
+
+class GracefulError(RuntimeError):
+    """Raised when encountering error that allows for a graceful shutdown."""
+    pass
```
erikrose: `pass` is just a hint for the lexer. You actually don't need it when you have anything else indented at this level, like your docstring.

danielhertenstein: I didn't know you could do that for a class definition. Very cool.

biancadanforth:
I'm not sure if this is all that you are doing, but you can also test CI stuff on a fork of this repo, so it doesn't e-mail folks who are watching the repo if you don't want to e-mail them.

cli/fathom_web/commands/vectorize.py
```diff
@@ -0,0 +1,312 @@
...
+            return new_file
+        except KeyError:
+            time.sleep(1)
+    else:
+        raise GracefulError(f'Could not find vectors file in {downloads_dir.as_posix()}.\n{vector_files_before} were already present.')
+
+
+def teardown(firefox, firefox_pid, geckoview_pid, server, server_thread, graceful_shutdown):
+    """
+    Close Firefox and the HTTP server.
```
biancadanforth: Should we also restore changes to `fathom-trainees`? (`git checkout -- .`) I'm just thinking about resetting the state of things at the end of the command. I don't think this particular task is going to break anything necessarily, it just modifies that repo and doesn't undo that modification.

cli/fathom_web/commands/vectorize.py
```diff
@@ -0,0 +1,312 @@
...
+
+    pages_text_area = firefox.find_element_by_id('pages')
+    pages_text_area.send_keys(sample_filenames)
+
+    downloads_dir = pathlib.Path(firefox.profile.default_preferences['browser.download.dir'])
+    vector_files_before = set(downloads_dir.glob('vector*.json'))
+    number_of_samples = len(sample_filenames.splitlines())
+    status_box = firefox.find_element_by_id('status')
+    vectorize_button = firefox.find_element_by_id('freeze')
+    completed_samples = 0
```
biancadanforth: We should also provide an option to select which `type` to train on (for example, the Fathom Smoot articles ruleset has both a `"paragraph"` and `"article"` type). Currently this will just vectorize whichever type is default selected in the dropdown.

cli/fathom_web/commands/vectorize.py
```diff
@@ -0,0 +1,312 @@
...
+        server.shutdown()
+        server_thread.join()
+    else:
+        # This is the only way I could get killing the program with ctrl+c to work properly.
+        if server:
+            server.shutdown()
+            server_thread.join()
+        if firefox:
+            try:
+                os.kill(firefox_pid, signal.CTRL_C_EVENT)
```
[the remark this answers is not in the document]
biancadanforth: I ran `fathom-vectorize` with the browser open to try and do some debugging, and there was no easy way to pause or stop the vectorization, so I closed the second window which triggered a [couple of errors](https://paste.mozilla.org/DWLYqJjb) -- one related to Selenium Webdriver: `selenium.common.exceptions.StaleElementReferenceException: Message: The element reference of <ul id="status"> is stale; either the element is no longer attached to the DOM, it is not in the current frame context, or the document has been refreshed`, and the other related to calling `teardown` in the `finally` block in `main`: `AttributeError: module 'signal' has no attribute 'CTRL_C_EVENT'`. I'm not sure about the first error, but the second definitely seems worth addressing.

Closing this window also terminated the HTTP server, so I couldn't manually load pages to debug. Maybe we should wait to shut down the HTTP server until after `firefox` has closed/been terminated?

cli/fathom_web/commands/vectorize.py
```diff
@@ -0,0 +1,312 @@
...
+    print('Configuring Firefox...', end='', flush=True)
+    options = webdriver.FirefoxOptions()
+    options.headless = not show_browser
+
+    profile = webdriver.FirefoxProfile()
+    profile.set_preference('browser.download.folderList', 2)
+    profile.set_preference('browser.download.dir', str(pathlib.Path(output_directory).absolute()))
+    profile.set_preference('browser.cache.disk.enable', False)
+    profile.set_preference('browser.cache.memory.enable', False)
+    profile.set_preference('browser.cache.offline.enable', False)
```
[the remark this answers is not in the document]
danielhertenstein: `browser.cache.disk.enable` will store pages on the hard disk. `browser.cache.memory.enable` will store pages in memory. `browser.cache.offline.enable` will store `<link rel="offline-resource">`s in some cache.

The disk version seems most important to me. The memory version seems debatable since each page is only loaded once during the `fathom-vectorize` run. The offline version is probably unnecessary, but I don't see a problem leaving it.

I don't know if there is a better source for what these do, but I used:
http://kb.mozillazine.org/Browser.cache.disk.enable
http://kb.mozillazine.org/Browser.cache.memory.enable
http://kb.mozillazine.org/Browser.cache.offline.enable

cli/fathom_web/commands/vectorize.py
```diff
@@ -0,0 +1,312 @@
...
+    for _ in range(5):
+        prefs = (pathlib.Path(firefox.capabilities.get('moz:profile')) / 'prefs.js').read_text().split(';')
+        uuids = next((line for line in prefs if 'extensions.webextensions.uuids' in line)).split(',')
+        try:
+            fathom_fox_uuid = next((line for line in uuids if 'fathomfox@mozilla.com' in line)).split('\\"')[3]
+            return fathom_fox_uuid
+        except StopIteration:
+            time.sleep(1)
+    else:
+        raise GracefulError('Could not find UUID for FathomFox. No entry in the pres.js file.')
```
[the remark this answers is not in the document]
danielhertenstein: I did think about generalizing this method when I wound up using it the second time, so I would support that.

cli/fathom_web/commands/vectorize.py
```diff
@@ -0,0 +1,312 @@
...
+    """
+    Look for errors in the vectorizer's status text.
+
+    If there is an error, raise an exception causing an ungraceful shutdown.
+    """
+    lines = status_text.splitlines()
+    for line in lines:
+        if 'failed:' in line:
+            return line
+    raise UngracefulError(f'There was a vectorizer error, but we could not find it in {status_text}')
```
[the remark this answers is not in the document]
danielhertenstein: I don't think it should be possible, but I want to raise an error here if it winds up being possible so the corner case can be discovered.

cli/fathom_web/commands/vectorize.py
```diff
@@ -0,0 +1,312 @@
...
+    """
+    for _ in range(5):
+        vector_files_after = set(downloads_dir.glob('vector*.json'))
+        try:
+            new_file = (vector_files_after - vector_files_before).pop()
+            return new_file
+        except KeyError:
+            time.sleep(1)
+    else:
+        raise GracefulError(f'Could not find vectors file in {downloads_dir.as_posix()}.\n{vector_files_before} were already present.')
```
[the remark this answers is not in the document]
danielhertenstein: I did not find any resources regarding file download events, but the browser-y portion of this is certainly not my expertise. I would love to learn such an event listener exists.

cli/fathom_web/commands/vectorize.py
```diff
@@ -0,0 +1,312 @@
...
+    SAMPLES_DIRECTORY: Path to the directory containing the sample pages
+    FATHOM_FOX_DIR: Path to the FathomFox source directory
+    FATHOM_TRAINEES_DIR: Path to the Fathom Trainees source directory
+
+    This tool will run an instance of Firefox to use the Vectorizer within the
+    FathomFox adddon. Required for this tool to work are:
+      * node
+      * yarn
+      * A FathomFox repository checkout
+      * A Fathom Trainees repository checkout
```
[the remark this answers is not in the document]
danielhertenstein: Yes, those are good points. I'll add them to the list, and those are additional configuration errors that need to have handling before this tool should be considered done.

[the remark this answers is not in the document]
danielhertenstein: And, yes, this uses whatever Firefox the OS has.

cli/fathom_web/commands/vectorize.py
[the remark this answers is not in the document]
danielhertenstein: I have not found a way to gracefully stop and close the Firefox windows while the vectorization is happening.

How had you paused or stopped the vectorizer before to do debugging?

If the user closes one of the windows, I think the entire vectorization should stop. What do you think?

I'll work on fixing the signal. I didn't realize until now that `CTRL_C_EVENT` was Windows only.

I'll also see if I can switch the order of the HTTP server closing and firefox closing. I recall needed to have this order before, but that reason may no longer be valid. Why are you trying to manually load pages during the vectorization?

cli/fathom_web/commands/vectorize.py
```diff
@@ -0,0 +1,312 @@
...
+    profile.set_preference('browser.download.dir', str(pathlib.Path(output_directory).absolute()))
+    profile.set_preference('browser.cache.disk.enable', False)
+    profile.set_preference('browser.cache.memory.enable', False)
+    profile.set_preference('browser.cache.offline.enable', False)
+
+    firefox = webdriver.Firefox(
+        options=options,
+        firefox_profile=profile,
+        service_log_path=f'{temp_dir.name}/geckodriver.log',
+    )
```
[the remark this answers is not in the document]
danielhertenstein: Yes, from the default installation.

cli/fathom_web/test/resources/vectorize/vectorize_ruleset.js
```diff
@@ -0,0 +1,56 @@
...
+let biases = [
+    ['secret', 0.0]
+];
+
+function caselessIncludes(haystack, needle) {
+    return haystack.toLowerCase().includes(needle.toLowerCase());
+}
+
+function hasSecretParagraph(fnode) {
+    console.log('this is running');
```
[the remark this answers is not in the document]
danielhertenstein: Thanks for catching this.

cli/fathom_web/commands/vectorize.py
```diff
@@ -0,0 +1,312 @@
...
+@argument('fathom_trainees_dir', type=Path(exists=True, file_okay=False))
+@option('--output-directory', '-o', type=Path(exists=True, file_okay=False), default=os.getcwd(),
+        help='Directory to save the vector file in (default: current working directory')
+@option('--show-browser', '-s', default=False, is_flag=True,
+        help='Flag to show browser window while running. Browser is run in headless mode by default.')
+def main(ruleset_file, samples_directory, fathom_fox_dir, fathom_trainees_dir, output_directory, show_browser):
+    """
+    Create feature vectors for a directory of training samples using a Fathom Ruleset.
+
+    RULESET_FILE: Path to the ruleset.js file
```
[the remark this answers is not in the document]
danielhertenstein: Good point.

cli/fathom_web/commands/vectorize.py
danielhertenstein: That is a very good point. Do you think it is important enough to do in this version of the tool or is an enhancement issue adequate?

cli/fathom_web/commands/vectorize.py
danielhertenstein: I agree. I hadn't thought about that. I'll add that for this PR.

cli/fathom_web/commands/vectorize.py
```diff
@@ -0,0 +1,312 @@
...
+    SAMPLES_DIRECTORY: Path to the directory containing the sample pages
+    FATHOM_FOX_DIR: Path to the FathomFox source directory
+    FATHOM_TRAINEES_DIR: Path to the Fathom Trainees source directory
```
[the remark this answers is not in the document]
danielhertenstein: I'll add the necessary formatting.

cli/fathom_web/commands/vectorize.py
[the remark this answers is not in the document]
danielhertenstein: > Firstly, since the error occurs after the vectorization has been attempted, that isn't an error during set up, is it? Maybe we should precede the error with "Error during setup" in this case.

I'll change the handling of `GracefulError` to remove reference to setup.
 
> Also, if we print an error that one of the pages failed during vectorization , do we need to still check and throw the `GracefulError` also? We probably shouldn't `look_for_new_vector_file` if the vectorizer throws an error, since it stops vectorizing in this case anyway.

Ah, I don't know how I didn't catch this one. We should be throwing a `UngracefulError` here because this is an error during vectorization and I haven't found a way to close everything during vectorization.

> Finally, in my case of first running this, `vector_files_before` is an empty set, so it just prints `set()` in the error message. Maybe if the set is empty, we just log that no vectors files were found in the download directory?

That is a good suggestion.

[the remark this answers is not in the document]
erikrose: If we do another iteration of this and program it against a proper API (rather than Selenium puppet-string-ing), we can just return the data itself rather than having the browser awkwardly lay it down on the filesystem. (Python can save it or pipe it, as the user demands.)

cli/fathom_web/commands/vectorize.py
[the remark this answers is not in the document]
biancadanforth: I don't think there's a great way other than looking at the blame in [SearchFox](https://searchfox.org/), which is a searchable directory of the tip of mozilla-central's source code, for where the pref is initialized and hoping you can track down the bug that introduced it in Bugzilla for more context. I didn't have much luck doing that though for at least the first pref.

I have never heard of MozillaZine! I noticed that was last updated in 2014, and was created by some community members. Could be a place to look, but I would take it with a grain of salt generally. I doubt these particular prefs have changed much if at all since 2014 in terms of what they control.

cli/fathom_web/commands/vectorize.py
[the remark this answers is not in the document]
biancadanforth: > How had you paused or stopped the vectorizer before to do debugging?

Typically in FathomFox, I just close the second window, and this would stop vectorization.

> If the user closes one of the windows, I think the entire vectorization should stop. What do you think?

Sounds like a good call, since one window has the pages we're vectorizing, and the other has the status of vectorization. If either closes, we can't really know that we successfully completed vectorization.

> Why are you trying to manually load pages during the vectorization?

I'm not trying to manually load pages during vectorization. I'm trying to pause vectorization and manually load pages in the browser between vectorization attempts.

Right now, FathomFox's `vectors.js` page lists the status, and for some errors, stops vectorization if it encounters an error. Some of these errors are tied to a particular page (e.g. `null` vectors will tell you which page(s) had `null` feature vector values), so maybe I want to open that page up and inspect it to fix the error.

One thing that helps (me) debug such errors (and helps with ruleset development in general) is to amend the `trainees.js` script, so that it actually runs the ruleset against the page as soon as its loaded into the page and logs the results. In this situation, I am using my fathom-trainees/fathom-fox vectorizer setup to load these pages outside of vectorization.

cli/fathom_web/commands/vectorize.py
biancadanforth: Yeah I think you could just add a TODO comment. As long as the `shutil.copyfile` command handles the situation where the file already exists, which it must, since `fathom-trainees` already comes with a default `trainees.js` file.

cli/fathom_web/commands/vectorize.py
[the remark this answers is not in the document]
biancadanforth: > I have not found a way to gracefully stop and close the Firefox windows while the vectorization is happening.

Note: We discussed this in IRC last Friday, when we learned that `driver.quit()` does not work due to this [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1430064).

cli/fathom_web/commands/vectorize.py
```diff
@@ -41,11 +41,13 @@ def main(ruleset_file, samples_directory, fathom_fox_dir, fathom_trainees_dir, o
     """
     Create feature vectors for a directory of training samples using a Fathom Ruleset.
 
-    RULESET_FILE: Path to the ruleset.js file
+    \b
```
biancadanforth: Huh; never seen that before. How is \b different from \n? They seem to behave the same when I swap them.

cli/fathom_web/commands/vectorize.py
```diff
@@ -54,6 +54,8 @@ def main(ruleset_file, samples_directory, fathom_fox_dir, fathom_trainees_dir, o
       * yarn
       * A FathomFox repository checkout
       * A Fathom Trainees repository checkout
+      * A copy of Firefox
+      * A geckoview driver downloaded and accessible in your PATH environment variable
```
biancadanforth: AFAIK, [GeckoView](https://wiki.mozilla.org/Mobile/GeckoView) is an Android library for Firefox. Can we just say "geckodriver"?

cli/fathom_web/commands/vectorize.py
```diff
@@ -285,7 +285,11 @@ def look_for_new_vector_file(downloads_dir, vector_files_before):
         except KeyError:
             time.sleep(1)
     else:
-        raise GracefulError(f'Could not find vectors file in {downloads_dir.as_posix()}.\n{vector_files_before} were already present.')
+        error_string = f'Cound not find vectors file in {downloads_dir.as_posix()}.'
```
biancadanforth: Nit: Typo s/Cound/Could

cli/fathom_web/test/resources/vectorize/vectorize_ruleset.js
```diff
@@ -0,0 +1,55 @@
...
+        biases
+    );
+}
+
+const trainees = new Map();
+const VIEWPORT_SIZE = {width: 1680, height: 950};
+
+const FEATURES = ['secret'];
+for (const feature of FEATURES) {
+    console.log(feature);
```
biancadanforth: Nit: console.log

cli/fathom_web/test/resources/vectorize/vectorize_ruleset.js
```diff
@@ -0,0 +1,55 @@
...
+        vectorType: feature,
+        rulesetMaker: () => makeRuleset(
+            [
+                ...coefficients.secret,
+            ],
+            biases
+        ),
+    };
+    trainees.set(feature, ruleset);
+    console.log(trainees);
```
biancadanforth: Nit: console.log

cli/fathom_web/commands/vectorize.py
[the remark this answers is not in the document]
biancadanforth: I don't see that this has been moved to a utilities function. Not blocking, but just want to remind you about it in case you want to do that before merging.

cli/fathom_web/commands/vectorize.py
```diff
@@ -0,0 +1,340 @@
...
+    downloads_dir = pathlib.Path(firefox.profile.default_preferences['browser.download.dir'])
+    vector_files_before = set(downloads_dir.glob('vector*.json'))
+    number_of_samples = len(sample_filenames.splitlines())
+    status_box = firefox.find_element_by_id('status')
+    vectorize_button = firefox.find_element_by_id('freeze')
+    completed_samples = 0
+    print('Done')
+
+    with progressbar(length=number_of_samples, label='Running Vectorizer...') as bar:
+        vectorize_button.click()
```
biancadanforth: When I run the Vectorizer with `-s`, sometimes the pages give the error that the server is taking too long to respond, and the connection timed out. Not all the time. I just ran this command `fathom-vectorize -s trainees.js article samples/training ~/Projects/fathom-fox/ ~/Projects/fathom-trainees/` twice, and the first time, it timed out with those errors, and the second time it vectorized successfully.

When it times out, I get the error: `Error: Invalid tab ID: X.` on the Vectorizer page.

Is it possible that the server sometimes drops some requests if they occur too quickly? This (particularly headless mode) is certainly the fastest we have ever requested these resources before from our HTTP server.

Perhaps we could file a follow up issue to investigate this more.

cli/fathom_web/test/test_vectorize.py
```diff
@@ -0,0 +1,25 @@
...
+
+    runner = CliRunner()
+    result = runner.invoke(
+        vectorize_main,
+        [
+            f'{test_dir}/resources/vectorize/vectorize_ruleset.js',
+            'secret',
+            f'{test_dir}/resources/vectorize/',
+            os.environ['FATHOM_FOX'],
+            os.environ['FATHOM_TRAINEES'],
```
biancadanforth: Unlike all our other `fathom-*` tests, this test [does not pass locally](https://paste.mozilla.org/EMVQBpw0) for me:
```
`E           KeyError: 'FATHOM_FOX'
../../.pyenv/versions/3.7.0/lib/python3.7/os.py:678: KeyError`
```
Though we use the environment variables in our CI, we don't define them when the test is run locally. I'm not sure the best way to solve this, as we don't know where the developer may have `fathom-fox` and `fathom-trainees` installed locally. What do you think? Also @erikrose ?

cli/fathom_web/commands/vectorize.py
```diff
@@ -0,0 +1,340 @@
...
+    """
+    Create .xpi files for fathom addons to load into Firefox.
+
+    The Firefox webdriver requires we load custom addons using .xpi files. We
+    need to load both FathomFox and Fathom Trainees. For Fathom Trainees, we
+    also need to run yarn to package up the addon with the user's ruleset.js.
+    """
+    print('Building fathom addons for Firefox...', end='', flush=True)
+    fathom_fox = create_xpi_for(pathlib.Path(fathom_fox_dir) / 'addon', 'fathom-fox', temp_dir)
+    shutil.copyfile(ruleset_file, f'{fathom_trainees_dir}/src/trainees.js')
```
biancadanforth: It doesn't seem that any errors are thrown for `fathom-vectorize` if the `ruleset.js` file passed in has `import` statements; although certainly when the script tries to load, it should throw an error in Firefox. Should we consider either checking in `fathom-vectorize` for `import` statements or checking in Fathom Fox and sending an error message back in this situation?

This would be a follow-up issue either way. Not blocking for this PR.

danielhertenstein:
Okay, the only thing left is the graceful shutdown and your note about debugging rulesets. We should talk about it after the holiday and decide if additional changes are necessary.

cli/fathom_web/commands/vectorize.py
danielhertenstein: \b will prevent line wrapping for the entire paragraph: https://click.palletsprojects.com/en/7.x/documentation/#preventing-rewrapping. For \n to work the same, I'd have to put it on many more lines.

cli/fathom_web/commands/vectorize.py
danielhertenstein: Whoops, you are right. Thanks!

cli/fathom_web/commands/vectorize.py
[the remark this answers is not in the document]
danielhertenstein: I have moved it. It feels only okay to me. Let me know what you think about it.

cli/fathom_web/commands/vectorize.py
danielhertenstein: Hmm, sadly, I haven't run into this. That could be. If you edit FathomFox to reduce the number of tabs (and reduce the speed of the requests), does the error go away?

cli/fathom_web/test/test_vectorize.py
danielhertenstein: I'm interested in both of your thoughts on needing to set this environment variable if you want to run the tests locally. I had the additional thought that `fathom-vectorize` could look for the environment variable and use them if they are present so the user doesn't have to specify the paths every time they use the tool.

cli/fathom_web/commands/vectorize.py
danielhertenstein: That is a good point. I lean towards letting Fathom Fox tell us what is wrong instead of doing up front validation.

cli/fathom_web/commands/vectorize.py
biancadanforth: I just want to mention that I have since encountered this problem outside of this PR with the latest FathomFox with a parallelized Vectorizer, so it’s definitely not related to the work here.

After trying your suggestion, I think there are two different bugs here: "Unable to connect" (an error page that Firefox displays in the tab where the page would have loaded that does not stop the Vectorizer) and "Error: Invalid tab ID: X" (an error that the extension throws that shows up in the status area of the Vectorizer page, and does stop the Vectorizer).

I modified FathomFox to use `maxTabs` for 8, 16 (default) and 24 leaving the default delay of 5s unchanged; here are the number of failures for each error:

|`maxTabs` | No issues  | Unable to connect | Invalid tab ID |
| ------------- | ------------- | ------------- | ------------- |
| 8 | 3  | 0  |  2 |
| 16 | 2  | 2  |  1 |
| 24 | 0  | 5  |  0 |

So thanks to your suggestion, I do think that my hunch was correct, and we may want to revisit our `maxTabs` value. I will file two separate issues for these.

cli/fathom_web/test/test_vectorize.py
biancadanforth: One option is, as part of the test, do what the CI does -- i.e. clone both repos and set them to the two environment variables -- just clean up at the end of the test. This may further help in situations where the user's local copies of either or both repos is not clean, and could cause the test to fail. I'm not sure then what the `.travis.yml` file would look like, however.

cli/fathom_web/commands/vectorize.py
biancadanforth: Okay; would you mind filing a follow-up issue for this? I imagine this will be a pretty common situation, and when I tried it, the error(s) were pretty opaque (nothing showed up in the Vectorizer page).

cli/fathom_web/test/test_vectorize.py
biancadanforth: We should agree on something here before merging this.

cli/fathom_web/commands/vectorize.py
```diff
@@ -0,0 +1,337 @@
...
+    print('Configuring Vectorizer...', end='', flush=True)
+    # Navigate to the vectorizer page
+    fathom_fox_uuid = get_fathom_fox_uuid(firefox)
+    firefox.get(f'moz-extension://{fathom_fox_uuid}/pages/vector.html')
+
+    pages_text_area = firefox.find_element_by_id('pages')
+    pages_text_area.send_keys(sample_filenames)
+
+    ruleset_dropdown_selector = Select(firefox.find_element_by_id('ruleset'))
+    ruleset_dropdown_selector.select_by_visible_text(fathom_type)
```
biancadanforth: When, for whatever reason FathomFox doesn’t find a `trainees.js`, it throws an [error](https://paste.mozilla.org/kiPcqCSS) of the form: `selenium.common.exceptions.NoSuchElementException: Message: Could not locate element with visible text: article`.

We should make this error message more helpful if possible as to the true issue, as I only learned it by using the `-s` option and briefly seeing the warning on the Vectorizer page before Firefox shut down ungracefully. This can be a follow-up issue.

cli/fathom_web/utils.py
```diff
@@ -62,7 +62,7 @@ def wait_for_function(function, error, max_tries):
     for _ in range(max_tries):
         try:
             return function()
-        except:  # noqa: E722
+        except Exception:  # noqa: E722
```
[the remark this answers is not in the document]
danielhertenstein: Before, this caught every type of exception. I have phrased this method such that the `function()` you are waiting on will throw an exception if it isn't "ready" yet. I had forgotten though that doing `except:` would catch a handful of exceptions that we're not actually interested in (like `KeyboardInterrupt`s). `except Exception` narrows the catching a little to set of exceptions we should except normal functions to throw: https://docs.python.org/3/library/exceptions.html#exception-hierarchy

cli/fathom_web/commands/vectorize.py
```diff
@@ -0,0 +1,337 @@
...
+        if server:
+            server.shutdown()
+            server_thread.join()
+        if firefox:
+            if os.name == 'nt':
+                signal_for_killing = signal.CTRL_C_EVENT
+            else:
+                signal_for_killing = signal.SIGTERM
+            try:
+                os.kill(firefox_pid, signal_for_killing)
```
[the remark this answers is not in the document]
danielhertenstein: This all sounds good to me. I'll add the note to the help message and open another issue.

danielhertenstein:
@biancadanforth, let's do some pair programming this afternoon to see what's happening with your trainees file. I have a guess.

cli/fathom_web/commands/vectorize.py
danielhertenstein: I agree.

cli/fathom_web/test/resources/vectorize/vectorize_ruleset.js
```diff
@@ -36,7 +36,6 @@ const VIEWPORT_SIZE = {width: 1680, height: 950};
 
 const FEATURES = ['secret'];
 for (const feature of FEATURES) {
-    console.log(feature);
```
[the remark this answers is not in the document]
danielhertenstein: We are not. I will add it.

cli/fathom_web/test/test_vectorize.py
danielhertenstein: If the checkout is done as part of `py_test` in the `Makefile` that could work. I'll mess around with that.

danielhertenstein: Okay, the real plan is to do this in the setup for the test in Python as was suggested by both @biancadanforth and @erikrose. I don't think I'll get this working before I go on leave though.

erikrose:
I am newly excited about this after introducing Erica to our tooling. You don't realize how much there is until you sit down and explain it to somebody. This tool could (or could evolve to) obsolete fathom-serve, yarn watch, and yarn browser: 3 fewer terminal tabs to keep track of. Plus it gets rid of the breaking-hard-link problem. It's on my short list of things to land once new-password ships.

danielhertenstein:
I agree. I'm still working on the Rust bug. I'm currently waiting on info from one of the interop people.

erikrose:
To do before landing:

- [ ] Change build_fathom_addons to compensate for FathomFox and Fathom and fathom-trainees repos merging.
- [ ] Replace run_file_server with fathom-serve or its guts.
- [ ] In run_vectorizer(), replace nice addon UUID (fathomfox@mozilla.com) with the 954efd86-8f62-49e7-8a65-80016051e382 one from the manifest.
- [ ] Teardown problems are minor ctrl-C stuff: may leave a headless Firefox and/or GeckoDriver running. Okay?
- [ ] If the vectorizer blows up with Invalid Tab ID stuff, fathom-vectorize will catch that and say so.
- [ ] Don't invoke fathom-list as a command; just call its guts.
- [ ] Look at TODOs.

erikrose:
Usable now. Before merging, add auto-vectorization to fathom-test, improve progress indication, and launch FF and build FathomFox only once for both sample sets.

PR #168, opened 2019-12-09
Require npm-installed tools to be around before we call them.

erikrose:
The rules that required $(JS) required .npm_installed indirectly, but I don't want to depend on that always holding true.


erikrose:
Rebased as 378c2ad1c2b34392b65c71b7eebf967e9cd4ec25.

PR #171, opened 2019-12-14
Add progress bar to fathom-extract.

erikrose:
Running this on fathom-login-forms took several minutes, so I wanted some feedback.


erikrose:
We decided to merge this because Bianca is inundated with reviews and the risk of this breaking anything is low.

PR #173, opened 2019-12-31
fathom-histogram

erikrose:
In the course of trying to smarten up the trainer and obsolete semi-manual binarization (cutoffs), I graphed histograms of all the feature values so I could get a sense of what the data shape is (and whether I should try normal-curve bucketing vs. quantile, etc.). I wrote it for myself, as a research tool, but it might also come in handy to show the distribution of new feature values, to see if they're what you thought they were going to be. Feel free to review, but feel equally free to ignore until it's proven itself useful.

You're probably best off reviewing the complete diff all at once.

What I found on the Shopping ruleset is a superlinear falloff on almost all features. In retrospect, this should be unsurprising for features that are almost all counts of things.

Screen Shot 2019-12-31 at 14 21 55

Btw, I don't think this actually has much overlap with Daniel's `rule_output_analysis.ipynb`.

### DIFF
--- a/cli/fathom_web/commands/histogram.py
+++ b/cli/fathom_web/commands/histogram.py
@@ -1,0 +1,24 @@
+from json import load
+
+from click import argument, command, File, option, progressbar
+from plotly.express import histogram
+from plotly.subplots import make_subplots
+
+from ..utils import classifier, tensor, tensors_from
+
+
+@command()
+@argument('vector_file', type=File('r'))
+@argument('feature', type=str)
+def main(vector_file, feature):
+    """Print a histogram of the given feature."""
+    data = load(vector_file)
+    x, y, num_yes = tensors_from(data['pages'])
+
+    feature_index = data['header']['featureNames'].index(feature)
+    feature_values = []
+    for page in data['pages']:
+        for node in page['nodes']:
+            feature_values.append(node['features'][feature_index])
+
+    histogram({feature: feature_values}, x=feature, nbins=40).show()
--- a/cli/fathom_web/commands/histogram.py
+++ b/cli/fathom_web/commands/histogram.py
@@ -7,21 +7,21 @@ from plotly.subplots import make_subplots
 
 @command()
 @argument('vector_file', type=File('r'))
-@option('--feature', '-f',
+@option('--feature_name', '-f',
         default='',
         help='The name of the feature to graph. Omit to graph all features.')
 @option('--show_zeros', '-z',
         is_flag=True,
         default=False,
         help='Show zero counts in the graphs. These tend to be large and dominate the scale of the graph, making it harder to see other bars.')
-def main(vector_file, feature, show_zeros):
+def main(vector_file, feature_name, show_zeros):
     """Print a histogram of one or more features."""
     data = load(vector_file)
     feature_names = data['header']['featureNames']
-    if feature:
-        feature_index = feature_names.index(feature)
+    if feature_name:
+        feature_index = feature_names.index(feature_name)
         vectors = [yes_and_no_vectors(data['pages'], feature_index, show_zeros)]
-        feature_names = [feature_names[feature_index]]
+        feature_names = [feature_name]
     else:
         vectors = [yes_and_no_vectors(data['pages'], feature_index, show_zeros)
                    for feature_index in range(len(feature_names))]
--- a/cli/fathom_web/commands/histogram.py
+++ b/cli/fathom_web/commands/histogram.py
@@ -4,8 +4,6 @@ from click import argument, command, File, option
 from plotly.graph_objects import Histogram
 from plotly.subplots import make_subplots
 
-from ..utils import tensors_from
-
 
 @command()
 @argument('vector_file', type=File('r'))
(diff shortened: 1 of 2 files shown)


cli/fathom_web/commands/histogram.py
```diff
@@ -0,0 +1,80 @@
...
+from click import argument, command, File, option
+from plotly.graph_objects import Histogram
+from plotly.subplots import make_subplots
+
+
+@command()
+@argument('vector_file', type=File('r'))
+@option('--feature_name', '-f',
+        default='',
+        help='The name of the feature to graph. Omit to graph all features.')
```
biancadanforth: If you can only pass one `vectors.json` file, is it possible to have more than one feature?

cli/fathom_web/commands/histogram.py
```diff
@@ -0,0 +1,80 @@
...
+        vectors = [yes_and_no_vectors(data['pages'], feature_index, show_zeros)
+                   for feature_index in range(len(feature_names))]
+    figure = histograms(vectors,
+                        feature_names,
+                        'Histograms of Feature Values' + ('' if show_zeros else
+                                                          ', Zeros Omitted'))
+    figure.show()
+
+
+def histograms(vectors, feature_names, title):
```
biancadanforth: Can we label the x- and y- axes in the plotly graphs? I imagine the x-axis is the raw feature out -- what is the y-axis? I guess I don't really know how to interpret these graphs for example:
Screen Shot 2020-01-08 at 8 55 37 AM

Also is it possible to associate a particular bar with a filename/sample name? E.g. maybe I'd like to know which negative sample has so much `textContent` in the `hasLongTextContent` feature's histogram.

cli/fathom_web/commands/histogram.py
biancadanforth: Ah by feature here we mean a "rule" in Fathom; I keep periodically confusing feature and class (or "type" in Fathom). So I guess I could graph just one feature (e.g. `hasLongTextContent` in my paragraph ruleset).

cli/fathom_web/commands/histogram.py
```diff
@@ -0,0 +1,80 @@
+from json import load
+
+from click import argument, command, File, option
+from plotly.graph_objects import Histogram
+from plotly.subplots import make_subplots
```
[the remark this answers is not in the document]
erikrose: Oops! I forgot to specify a dependency. Good review! :-)

cli/fathom_web/commands/histogram.py
erikrose: I should say "type" for consistency.

cli/fathom_web/commands/histogram.py
erikrose: Labeling the axes is a good call. I'll see if it makes things ridiculously noisy, in which case I'll look for a central place to convey the information.

Since these are histograms, each bar is an adding-up of potentially many samples. But maybe we could put the n-heaviest in a tooltip.

erikrose: Self: All 3 of the above features in the graph should be seeing all <p> elements. Why do some see a total of hundreds and the others only 20?

biancadanforth: We just discussed in our 1:1 that the y-axis should be the number of candidate nodes found for each feature.

Why does the `hasEllipsisAtEndOfText` only have 20 candidate nodes (18 negative and 2 positive) while hasLongTextContent has 350+ candidate nodes? All of these rules are run without `when` clauses. Shouldn’t they all have the same number of candidate nodes?

Here's the ruleset I was using (I was only using a `vectors.json` for the paragraph ruleset):
```js
function makeRuleset(coeffs, biases) {
  return ruleset([
    /**
      * Paragraph rules
    */
    // Consider all visible paragraph-ish elements
    rule(dom("p, pre").when(isElementVisible), type("paragraph")),
    rule(dom("div").when(isElementVisible).when(divHasBrChildElement), type("paragraph").note(numParagraphTextNodesInDiv)),
    rule(dom("div").when(isElementVisible).when(divHasOnlyTextNodesAnchorElementsOrSpanElements), type("paragraph").note(numParagraphTextNodesInDiv)),
    rule(type("paragraph"), score(pElementHasListItemAncestor), {name: "pElementHasListItemAncestor"}),
    rule(type("paragraph"), score(hasLongTextContent), {name: "hasLongTextContent"}),
    rule(type("paragraph"), score(containsElipsisAtEndOfText), {name: "containsElipsisAtEndOfText"}),
    rule(type("paragraph"), score(classNameIncludesCaption), {name: "classNameIncludesCaption"}),
    // return paragraph-ish element(s) with max score
    rule(type("paragraph").max(), out("paragraph")),

    /**
      * Article rules
    */
    rule(dom("html"), type("article")),
    rule(type("article"), score(hasEnoughParagraphs), {name: "hasEnoughParagraphs"}),
    rule(type("article"), score(hasExactlyOneArticleElement), {name: "hasExactlyOneArticleElement"}),
    rule(type("article"), score(paragraphElementsHaveSiblingsWithSameTagName), {name: "paragraphElementsHaveSiblingsWithSameTagName"}),
    rule(type("article"), score(mostParagraphElementsAreHorizontallyAligned), {name: "mostParagraphElementsAreHorizontallyAligned"}),
    rule(type("article"), score(moreParagraphElementsThanListItemsOrTableRows), {name: "moreParagraphElementsThanListItemsOrTableRows"}),
    rule(type("article"), score(headerElementIsSiblingToParagraphElements), {name: "headerElementIsSiblingToParagraphElements"}),
    rule(type("article"), score(hasMultipleArticleElements), {name: "hasMultipleArticleElements"}),
    rule(type("article"), score(hasMultipleParagraphsWhoseClassNameIncludesArticle), {name: "hasMultipleParagraphsWhoseClassNameIncludesArticle"}),
    rule(type("article"), out("article"))
  ],
  coeffs,
  biases);
}
```

erikrose:
Wrote a tty-based one and shipped it. We'll go with that for now. Though note #242 as a possible future enhancement.

PR #183, opened 2020-03-05
Graph the timings that FathomFox now records.

erikrose:
(Needs the same-named branch on FathomFox to emit them.)


danielhertenstein:
You need to add sklearn to the requirements.

erikrose:
Thanks; good catch! How about now?

PR #184, opened 2020-03-06
Teach fathom-list how to optionally spit out original URLs of samples.

erikrose:
I needed this a second time, so I guess it's time to make it publicly available!

Specifically, I'm using it to tell which domains were assigned to Vlad and actually yielded a sample, so I don't assign him the same domain a second time.


erikrose:
Thanks!

PR #186, opened 2020-03-07
Also report the average time per tag, because that's more relevant for some applications.

erikrose:
A feel-good project for the end of a Friday night


cli/fathom_web/commands/train.py
```diff
@@ -8,7 +8,7 @@
 from torch.optim import Adam
 
 from ..accuracy import accuracy_per_tag, accuracy_per_page, pretty_accuracy
-from ..utils import classifier, mini_histogram, tensors_from
+from ..utils import classifier, mini_histogram, speed_readout, tensors_from
```
danielhertenstein: `mini_histogram` is no longer used here.

erikrose: Thanks! In test.py, too, it turns out.

PR #187, opened 2020-03-12
Per tag report

erikrose:
Convert the per-page table into a per-tag one. This should help us chase down failures better. This needs the FathomFox branch of the same name if you want to see the markup snippets in the table.

### DIFF
--- a/cli/fathom_web/accuracy.py
+++ b/cli/fathom_web/accuracy.py
@@ -28,6 +28,11 @@ def accuracy_per_tag(y, y_pred):
         return (successes / number_of_tags), false_positives, false_negatives
 
 
+# Max chars from a start tag we're willing to display in the per-tag metrics
+# report:
+MARKUP_MAX_WIDTH = 36
+
+
 def per_tag_metrics(page, model):
     """Return the per-tag numbers to be templated into a human-readable report
     by ``print_per_tag_report``."""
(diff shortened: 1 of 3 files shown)


danielhertenstein:
The tests in `test.py` are broken since `accuracy_per_page()` no longer exists.

cli/fathom_web/accuracy.py
```diff
@@ -28,76 +28,104 @@ def accuracy_per_tag(y, y_pred):
         return (successes / number_of_tags), false_positives, false_negatives
 
 
+# Max chars from a start tag we're willing to display in the per-tag metrics
+# report:
+MARKUP_MAX_WIDTH = 36
```
danielhertenstein: Since you're returning a slice of max size 36 in the [FathomFox PR](https://github.com/mozilla/fathom-fox/pull/75/files#diff-ce5ff8cbb3b5f2aa37a784e7132e2a95R122l), do you need this constant here?

erikrose: It's belt and suspenders. As I said in 1cccb87fe3d90531cdda81586ef560d1b6a72e8d's commit message, "the exact length may change someday, and I don't want the layout to break".

erikrose:
Yay! :-D

PR #189, opened 2020-03-12
Updates training metrics

danielhertenstein:
Changes false positive ratio and false negative ratio to rates and adds f1 score. I believe this to be more in line with the standard metrics used in machine learning.

### DIFF
--- a/cli/fathom_web/accuracy.py
+++ b/cli/fathom_web/accuracy.py
@@ -150,19 +150,23 @@ def pretty_accuracy(description, accuracy, number_of_samples, false_positives=No
     """
     ci_low, ci_high = confidence_interval(accuracy, number_of_samples)
     if false_positives is not None:
-        false_positive_ratio = false_positives / number_of_samples
-        false_negative_ratio = false_negatives / number_of_samples
-        fp_ci_low, fp_ci_high = confidence_interval(false_positive_ratio, positives)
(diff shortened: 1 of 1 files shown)


cli/fathom_web/accuracy.py
```diff
@@ -150,19 +150,23 @@ def pretty_accuracy(description, accuracy, number_of_samples, false_positives=No
...
         precision = true_positives / (true_positives + false_positives)
-        recall = true_positives / positives
+        # recall is the same as the true positive rate
+        recall = 1 - false_negative_rate
+        f1score = 2.0 * (precision * recall) / (precision + recall)
         falses = ('\n'
-                  f'                         FP:  {false_positive_ratio:.5f}    95% CI: ({fp_ci_low:.5f}, {fp_ci_high:.5f})\n'
-                  f'                         FN:  {false_negative_ratio:.5f}    95% CI: ({fn_ci_low:.5f}, {fn_ci_high:.5f})\n'
-                  f'                  Precision:  {precision:.5f}    Recall: {recall:.5f}\n')
+                  f'                        FPR:  {false_positive_rate:.5f}    95% CI: ({fpr_ci_low:.5f}, {fpr_ci_high:.5f})\n'
```
erikrose: I'm so glad you changed the labels so we can tell when we're comparing two unalike statistics.

danielhertenstein:
I agree that the terminology for this one does not seem the clearest. And you make a good point that we should be careful about how we phrase our metrics so the denominator is clear.

PR #193, opened 2020-03-13
Add ability to tune decision threshold

danielhertenstein:
Adds a decision threshold option to `fathom-train` and an argument to `fathom-test`. I added it as an argument to `fathom-test` so the user will be less inclined to forget they used some custom threshold in train.

### DIFF
--- a/cli/fathom_web/commands/test.py
+++ b/cli/fathom_web/commands/test.py
@@ -47,8 +47,10 @@ def model_from_json(weights, num_outputs, feature_names):
 @command()
 @argument('testing_file',
           type=File('r'))
+@argument('decision-threshold',
+          type=float)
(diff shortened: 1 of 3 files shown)


cli/fathom_web/commands/test.py
```diff
@@ -47,8 +47,10 @@ def model_from_json(weights, num_outputs, feature_names):
 @command()
 @argument('testing_file',
           type=File('r'))
+@argument('decision-threshold',
```
erikrose: What about calling it "confidence", since I think we call it that everywhere else. Or "score". But "decision" is new.

danielhertenstein: That's a good point. I'll go with confidence.

PR #196, opened 2020-03-17
Add `againstElement()`, which lets us efficiently classify only a single element.

erikrose:
Considered making a workalike to `dom()` called `element()` or `one()` (and aliasing `dom()` to `all()`), but then you'd need to change a ruleset every time you wanted to use it differently. That should be the application's decision, not the ruleset writer's. And, of immediate practical value, any ruleset using `element()` in production would need a clumsy switch to use `dom()` in training.

### DIFF
--- a/docs/ruleset.rst
+++ b/docs/ruleset.rst
@@ -10,7 +10,7 @@ The top-level Fathom object is the ruleset, an unordered collection of rules. Th
 .. autofunction:: ruleset
 
 .. autoclass:: Ruleset
-   :members: against, againstElement, rules
+   :members: against, rules
 
 Then you call :func:`Ruleset.against` to get back a :class:`BoundRuleset`, which is specific to a given DOM tree. From that, you pull answers.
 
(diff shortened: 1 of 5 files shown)


erikrose:
I can't think of any situations where this global transformation of `dom()` calls into `element()` calls would be a problem, even in multi-type rulesets. Can you?

danielhertenstein:
I think if you had a ruleset like:
```
rule(dom("a"), type("bar")),
rule(type("bar", score(someScoringMethod)),
rule(type("bar").max(), out("bar")),

rule(dom("input"), type("foo")),
rule(type("foo"), score(doesSomethingWithBars), 
rule(type("foo"), out("foo"))
```
where you used the anchors in a page to make a inform your decision about each input element, and you called `againstElement()` for some input, you wouldn't get the anchors you expected.

erikrose:
Good point. Okay, another way we could do it is to go back to the `element()` approach but rig FathomFox to implicitly make `element()` calls behave like `dom()` ones during training. Sound good?

danielhertenstein:
I think that's reasonable.

erikrose:
To do: rig up a switch so FathomFox can make element() calls turn into dom() ones during vectorization and evaluation. But that can be a separate PR.

PR #199, opened 2020-03-20
Add an `--exclude`` option to the trainer to help with ablation.

erikrose:


cli/fathom_web/utils.py
```diff
@@ -17,26 +17,45 @@ def tensor(some_list):
...
+
     """
+    if exclude is None:
+        exclude = []
+    excluded_indices = [vector_data['header']['featureNames'].index(e) for e in exclude]
+
     xs = []
     ys = []
     num_targets = 0
+    pages = vector_data['pages']
```
danielhertenstein: I really like this little change.

erikrose: Thanks! I had little choice; I needed to start passing the whole vector file in here instead of just the pages, and I wasn't about to say `vector_data['pages']` 3 times. :-)

danielhertenstein:
The list of features printed out after training is wrong. It removes the last N depending on how many features you exclude. It does look like the right features are being trained, but the output of the coefficients does not match those features.

erikrose:
Ah yes. The per-tag metrics are probably wrong, too.

erikrose:
Squashed and merged in 89a4219a934b01e111c83f191239528c78eca2b9.

PR #203, opened 2020-03-24
Merge FathomFox into the Fathom repo

erikrose:
Fixes https://github.com/mozilla/fathom-fox/issues/66.

There's now a top-level makefile with shortcuts to common tasks, but the real meat and potatoes are in the fathom and cli folders, each of which contains its own Makefile. Later I want to hook up fathom_fox similarly, but, for now, it still uses its old yarn scripts.

I left some of my Travis-debugging commits in, because I want them as reference when I someday go solve the venv problem properly. For now, I am unable to get Travis to use the venvs our build process creates; instead, it insists on using the one Travis implicitly provides. Calling `python3 -m venv venv` on Travis results in a venv without a copy of pip installed, which makes it hard to get any further.

After merging:
- [ ] Transfer FathomFox issues to this repo.
- [ ] Archive the FathomFox repo.
- [ ] Re-point any links in readmes.
- [ ] Get linting to work in FathomFox, and fire it off from the top-level Makefile.
- [ ] Replace yarn scripts in FathomFox with a makefile.


erikrose:
To review this, I think your best bet is to look at the overall diff and ignore anything in the fathom_fox folder. I didn't change anything about FathomFox except to move it into that folder.

erikrose:
The CI build info doesn't seem to be showing up on this page. Here it is: https://travis-ci.org/github/mozilla/fathom/builds/666500596.

docs/conf.py
```diff
@@ -37,6 +37,8 @@
     'sphinx_js'
 ]
 
+js_source_path = '../fathom'
```
danielhertenstein: Running `make docs` gives me this error:
```
Running Sphinx v2.4.4
making output directory... done
ERROR: Unable to find the source file or directory C:\cygdrive\c\Users\Daniel\code\erik-fathom\fathom

Sphinx error:
jsdoc found no JS files in the directories ['/cygdrive/c/Users/Daniel/code/erik-fathom/fathom']. Make sure js_source_path is set correctly in conf.py. It is also possible (though unlikely) that jsdoc emitted invalid JSON.
make[1]: *** [Makefile:75: html] Error 2
make[1]: Leaving directory '/cygdrive/c/Users/Daniel/code/erik-fathom/docs'
make: *** [Makefile:4: docs] Error 2
```
This is some Windows/cygwin thing that also happened before with the venv. I didn't find an easy solution yet, but I will continue looking.

erikrose: I'm curious whether running `make clean html` within the docs folder acts any differently.

erikrose: You might try `VIRTUAL_ENV ?= ./venv` instead of `VIRTUAL_ENV ?= venv` in docs/Makefile. That's the only difference I see between the docs makefile on this branch and on master.

erikrose: Of course, there's also the fact that I updated to the latest Sphinx, sphinx-js, etc., etc. Now I wish I hadn't done that.

danielhertenstein: You know what, I didn't have the right base virtual environment when trying to build the docs the first time (I discovered this when the cli virtual environment wasn't building right) and running `make clean` does not remove the virtual environment in the docs directory so I still had a malconfigured virtual environment when trying to investigate. I manually deleted the first virtual environment and it works great now that I have the correct base virtual environment making fathom's virtual environments.

danielhertenstein:
Although, I realize that I just left a comment saying the virtual environment in docs wasn't removed when running `make clean`, but then immediately approved this. Assuming the `clean` target in the docs Makefile is changed to remove the virtual environment, I'm happy with this :)

erikrose:
I had it that way on purpose, since it's Sphinx's usual procedure to run `make clean` before a build, which would cause a pricey venv reinstall. But I'll fiddle with things until there's a separate target for cleaning just the built docs.

danielhertenstein:
Ah, I see the advantage now. In case you haven't already seen it, the last commit included `venv/bin/activate` which I don't think should be there.

erikrose:
Thanks! I missed that!

erikrose:
Merged in 511a6d4fc21687e8aaf28a1591dea845ca959696.

PR #243, opened 2020-06-29
Put quotes around PATH in cli makefile to make Windows happy

danielhertenstein:


danielhertenstein:
Sounds good. I opened it as a draft because I wanted to see if it passed the CI before I bothered you with it.

PR #244, opened 2020-06-30
Adds changes needed to get vectorization to "work" on Windows.

danielhertenstein:
I do run into an error every time during vectorization though. Regardless of how many samples, threads, delay, I always hit a `Error: can't access property "browser", tab is null` during vectorization. I don't think I've hit this before. I want to get this over to @erikrose to see if the changes I made to get autovec working on Windows cause the same problems on Mac.

### DIFF
--- a/cli/fathom_web/vectorizer.py
+++ b/cli/fathom_web/vectorizer.py
@@ -194,7 +194,7 @@ def fathom_fox_addon(ruleset_file):
 
 def add_cmd_if_windows(binary_path):
     """Add the .cmd suffix to a pathlib.Path if running on Windows."""
-    if platform.system() == 'Windows':
+    if os.name == 'nt':
         return binary_path.with_suffix('.cmd')
     else:
         return binary_path
@@ -284,12 +284,14 @@ def locked_cached_fathom():
             # Install yarn, because it installs FathomFox's dependencies in 15s rather
             # than 30s. And once installed once, yarn itself takes only 1.5s to install
             # again. Also, it has security advantages. Though this install itself isn't
-            # hashed, so we're just trusting NPM.
+            # hashed, so we're just trusting NPM. Additionally, we have to call
+            # shutils.which() on npm because it is a .cmd file on Windows, while node
+            # (down below) is an executable.
             run_in_fathom_fox(which('npm'), 'install', 'yarn@1.22.4',
                               desc='Installing yarn')
 
             # Figure out how to invoke yarn:
-            if platform.system() == 'Windows':
+            if os.name == 'nt':
                 yarn_binary = str((fathom_fox / 'node_modules' / 'yarn' / 'bin' / 'yarn.js').resolve())
             else:
                 yarn_binary = str((fathom_fox / 'node_modules' / '.bin' / 'yarn').resolve())
--- a/cli/fathom_web/vectorizer.py
+++ b/cli/fathom_web/vectorizer.py
@@ -174,11 +174,7 @@ def fathom_fox_addon(ruleset_file):
                 copyfileobj(ruleset_file, new_ruleset_file)
 
             # Build FathomFox:
-            if platform.system() == 'Windows':
-                rollup_path = str((fathom_fox / 'node_modules' / '.bin' / 'rollup.cmd').resolve())
-            else:
-                rollup_path = str((fathom_fox / 'node_modules' / '.bin' / 'rollup').resolve())
-            run(rollup_path,
+            run(str(add_cmd_if_windows((fathom_fox / 'node_modules' / '.bin' / 'rollup').resolve())),
                 '-c',
                 cwd=fathom_fox,
                 desc='Compiling ruleset')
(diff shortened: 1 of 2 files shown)


danielhertenstein:
Interestingly this now fails on Linux while working on Windows. I'll get WSL put together so I can dig into it.

cli/fathom_web/vectorizer.py
```diff
@@ -175,7 +174,7 @@ def fathom_fox_addon(ruleset_file):
                 copyfileobj(ruleset_file, new_ruleset_file)
 
             # Build FathomFox:
-            run(fathom_fox / 'node_modules' / '.bin' / 'rollup',
+            run(str((fathom_fox / 'node_modules' / '.bin' / 'rollup').resolve()),
```
erikrose, line 177: We say `str((…).resolve())` a bunch of times now. Seems like we should factor it out.

erikrose:
Yes, it fails on the Mac too. The Ruleset menu on the Vectorizer page is empty. Autovectorization reports `Configuring Vectorizer...Error: Couldn't find trainee ID "body" in your rulesets.`, but I can see it more readily by running `fathom-fox`.

erikrose:
Getting rid of `shell=True` fixes it.

erikrose:
The breakage is during the call to rollup.

erikrose:
Here's the problem: when you set `shell=True`, `args` is expected to be a string, not an array of args. So, changing the line to this works:

`return subprocess.run(' '.join(args), cwd=cwd, shell=True, capture_output=True, check=True)`

Of course, that would require some quoting to be safe, and quoting is not my favorite. Is the need for the shell due to the need to interpret a `.cmd` file?

danielhertenstein:
I'm looking into the need of `shell` right now.

erikrose:
I wonder if you could explicitly say `cmd geckodriver.cmd` or something.

cli/fathom_web/vectorizer.py
```diff
@@ -175,7 +174,11 @@ def fathom_fox_addon(ruleset_file):
                 copyfileobj(ruleset_file, new_ruleset_file)
 
             # Build FathomFox:
-            run(fathom_fox / 'node_modules' / '.bin' / 'rollup',
+            if platform.system() == 'Windows':
```
erikrose, line 177: We have this "cmd-or-not" block repeat a few times. It might be a good thing to factor up.

danielhertenstein:
One of the three cases requires getting the JS file instead of the `.cmd`, so I only addressed the other two in order to keep the factored out function simple.

cli/fathom_web/vectorizer.py
```diff
@@ -190,7 +189,15 @@ def fathom_fox_addon(ruleset_file):
         # It should be okay to reference geckodriver outside the
         # locked_cached_fathom() lock, since that part of the cache is
         # immutable:
-        yield addon_path, (fathom_fox / 'node_modules' / '.bin' / 'geckodriver')
+        yield addon_path, add_cmd_if_windows(fathom_fox / 'node_modules' / '.bin' / 'geckodriver')
+
+
+def add_cmd_if_windows(binary_path):
+    """Add the .cmd suffix to a pathlib.Path if running on Windows."""
+    if platform.system() == 'Windows':
```
erikrose, line 197: You're still running Cygwin, correct? And you still get `.cmd` files laid down? (I ask in case we should instead be checking for `sys.platform == 'win32'`.)

cli/fathom_web/vectorizer.py
```diff
@@ -278,30 +285,12 @@ def run_in_fathom_fox(*args, desc):
             # than 30s. And once installed once, yarn itself takes only 1.5s to install
             # again. Also, it has security advantages. Though this install itself isn't
             # hashed, so we're just trusting NPM.
-            run_in_fathom_fox('npm', 'install', 'yarn@1.22.4',
+            run_in_fathom_fox(which('npm'), 'install', 'yarn@1.22.4',
```
erikrose, line 288: It's funny that this required `which()` but the call to `node` below didn't. Is that on purpose?

cli/fathom_web/vectorizer.py
danielhertenstein: I do have Cygwin around, but I'm not in a Cygwin terminal to run this (or anything). It's mostly there to provide the utilities the Makefile uses. Regardless, I think you're right that we should be checking for `sys.platform`.

cli/fathom_web/vectorizer.py
danielhertenstein: It is funny. I believe the binaries `npm` and `node` are in the same place, so I would think the `which()` would be required for both or none. I'll check that there isn't anything weird going on with another installation someplace I don't know about.

danielhertenstein: Ah, `node` is an actual executable while `npm` is a `.cmd`.

erikrose: That makes sense, then. Worth a comment so we don't have to wonder about it again.

PR #246, opened 2020-07-02
Update docs after looking at them with "fresh eyes" (I was away for 3 months).

danielhertenstein:
Particular things I'd like feedback on (in addition to any and all other feedback, of course):
- Should the Windows Considerations section be lower on the page?
- Is the code block in Getting Ready to Train too large? Should I take out some of the comments?
- Is the Evaluator section too large of a block of text? Should it have any screencaptures?

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -7,15 +7,10 @@ Source
 
 It's on `GitHub <https://github.com/mozilla/fathom>`_.
 
-Windows Considerations
-======================
-
-Fathom uses `Makefiles <https://www.gnu.org/software/make/manual/make.html>`_ to build its npm package, commandline tools, and docs; as well as run its tests. Additionally, these Makefiles rely on Unix commands. Therefore, if you are developing on Windows, you need access to these Unix commands through something like `Cygwin <https://www.cygwin.com/>`_. You can build and test Fathom using `Windows Subsystem for Linux <https://docs.microsoft.com/en-us/windows/wsl/>`_, but just know that you are technically building and testing Fathom in Linux when you do.
-
 Tests and Examples
 ==================
 
-To run `the tests <https://github.com/mozilla/fathom/tree/master/fathom/test>`_, run... ::
+To run the tests, run... ::
 
     make lint test
 
@@ -42,3 +37,8 @@ Gotchas
 =======
 
 If you are developing the CLI tools and your changes to their embedded copy of the Fathom JS framework don't seem to be taking effect, commit first. The make target that builds ``fathom.zip`` uses ``git archive`` to pull from ``HEAD``. In this scenario, we tend to use a single local commit we amend with ``git commit --amend --no-edit`` when we want to test our changes.
+
+Windows Considerations
+======================
+
+Fathom uses `Makefiles <https://www.gnu.org/software/make/manual/make.html>`_ to do its builds and run its tests. These Makefiles rely on Unix commands. Therefore, if you are developing on Windows, you need access to these Unix commands through something like `Cygwin <https://www.cygwin.com/>`_. You can build and test Fathom using `Windows Subsystem for Linux <https://docs.microsoft.com/en-us/windows/wsl/>`_, but just know that you are technically building and testing Fathom in Linux when you do.
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -7,12 +7,15 @@ Source
 
 It's on `GitHub <https://github.com/mozilla/fathom>`_.
 
+Windows Considerations
+======================
+
+Fathom uses `Makefiles <https://www.gnu.org/software/make/manual/make.html>`_ to build its npm package, commandline tools, and docs; as well as run its tests. Additionally, these Makefiles rely on Unix commands. Therefore, if you are developing on Windows, you need access to these Unix commands through something like `Cygwin <https://www.cygwin.com/>`_. You can build and test Fathom using `Windows Subsystem for Linux <https://docs.microsoft.com/en-us/windows/wsl/>`_, but just know that you are technically building and testing Fathom in Linux when you do.
+
 Tests and Examples
 ==================
 
-`Our tests <https://github.com/mozilla/fathom/tree/master/fathom/test>`_ are replete with examples exercising every corner of Fathom.
-
-To run the tests, run... ::
+To run `the tests <https://github.com/mozilla/fathom/tree/master/fathom/test>`_, run... ::
 
     make lint test
 
@@ -38,4 +41,4 @@ To build the docs... ::
 Gotchas
 =======
 
-If you are developing the CLI tools and your changes to their embedded copy of the Fathom JS framework don't seem to be taking effect, commit first. The make target that builds ``fathom.zip`` uses ``git archive`` to pull from ``HEAD``.
+If you are developing the CLI tools and your changes to their embedded copy of the Fathom JS framework don't seem to be taking effect, commit first. The make target that builds ``fathom.zip`` uses ``git archive`` to pull from ``HEAD``. In this scenario, we tend to use a single local commit we amend with ``git commit --amend --no-edit`` when we want to test our changes.
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -41,4 +41,4 @@ If you are developing the CLI tools and your changes to their embedded copy of t
 Windows Considerations
 ======================
 
-Fathom uses `Makefiles <https://www.gnu.org/software/make/manual/make.html>`_ to do its builds and run its tests. These Makefiles rely on Unix commands. Therefore, if you are developing on Windows, you need access to these Unix commands through something like `Cygwin <https://www.cygwin.com/>`_. You can build and test Fathom using `Windows Subsystem for Linux <https://docs.microsoft.com/en-us/windows/wsl/>`_, but just know that you are technically building and testing Fathom in Linux when you do.
+Fathom uses `makefiles <https://www.gnu.org/software/make/manual/make.html>`_ to do its builds and run its tests. These makefiles rely on Unix commands. Therefore, if you are developing on Windows, you need access to these Unix commands through something like `Cygwin <https://www.cygwin.com/>`_. You can build and test Fathom using `Windows Subsystem for Linux <https://docs.microsoft.com/en-us/windows/wsl/>`_, but just know that you are technically building and testing Fathom in Linux when you do.
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -6,6 +6,17 @@ Fathom consists of 3 parts. Here's how to install each one.
 
 .. _fathomfox-installation:
 
+Commandline Tools
+=================
+
+Fathom's commandline tools take your labeled pages as input and train the machine-learning model. They also contain the simplest way to collect pages, FathomFox. If you don't already have Python 3.7 or better, download it from https://www.python.org/downloads/. Then, install the tools by running... ::
+
+    pip3 install fathom-web
+
+It's possible your Python package manager is called simply "pip" rather than "pip3". Give that a try if the above fails.
+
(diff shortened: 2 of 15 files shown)


docs/development.rst
```diff
@@ -7,12 +7,15 @@ Source
...
+
+Fathom uses `Makefiles <https://www.gnu.org/software/make/manual/make.html>`_ to build its npm package, commandline tools, and docs; as well as run its tests. Additionally, these Makefiles rely on Unix commands. Therefore, if you are developing on Windows, you need access to these Unix commands through something like `Cygwin <https://www.cygwin.com/>`_. You can build and test Fathom using `Windows Subsystem for Linux <https://docs.microsoft.com/en-us/windows/wsl/>`_, but just know that you are technically building and testing Fathom in Linux when you do.
+
 Tests and Examples
 ==================
 
-`Our tests <https://github.com/mozilla/fathom/tree/master/fathom/test>`_ are replete with examples exercising every corner of Fathom.
-
-To run the tests, run... ::
+To run `the tests <https://github.com/mozilla/fathom/tree/master/fathom/test>`_, run... ::
```
erikrose, line 18: Maybe just forget about the link. We have 2 separate test folders, after all.

docs/development.rst
```diff
@@ -7,12 +7,15 @@ Source
 
 It's on `GitHub <https://github.com/mozilla/fathom>`_.
 
+Windows Considerations
+======================
+
+Fathom uses `Makefiles <https://www.gnu.org/software/make/manual/make.html>`_ to build its npm package, commandline tools, and docs; as well as run its tests. Additionally, these Makefiles rely on Unix commands. Therefore, if you are developing on Windows, you need access to these Unix commands through something like `Cygwin <https://www.cygwin.com/>`_. You can build and test Fathom using `Windows Subsystem for Linux <https://docs.microsoft.com/en-us/windows/wsl/>`_, but just know that you are technically building and testing Fathom in Linux when you do.
```
erikrose, line 13: I think we could tersen: "Fathom uses makefiles to do its builds and run its tests. These rely on Unix commands. ..."

docs/development.rst
```diff
@@ -38,4 +41,4 @@ To build the docs... ::
 Gotchas
 =======
 
-If you are developing the CLI tools and your changes to their embedded copy of the Fathom JS framework don't seem to be taking effect, commit first. The make target that builds ``fathom.zip`` uses ``git archive`` to pull from ``HEAD``.
+If you are developing the CLI tools and your changes to their embedded copy of the Fathom JS framework don't seem to be taking effect, commit first. The make target that builds ``fathom.zip`` uses ``git archive`` to pull from ``HEAD``. In this scenario, we tend to use a single local commit we amend with ``git commit --amend --no-edit`` when we want to test our changes.
```
erikrose, line 44: thumbs up!

docs/development.rst
erikrose: And yes, we could probably move the Windows section to the bottom of the page, as it'll likely be used seldom.

docs/installing.rst
```diff
@@ -6,6 +6,17 @@ Fathom consists of 3 parts. Here's how to install each one.
 
 .. _fathomfox-installation:
 
+Commandline Tools
+=================
+
+Fathom's commandline tools take your labeled pages as input and train the machine-learning model. They also contain the simplest way to collect pages, FathomFox. If you don't already have Python 3.7 or better, download it from https://www.python.org/downloads/. Then, install the tools by running... ::
```
erikrose, line 12: "Contain" a "way" sounds funny to my ear. Maybe we could just say "They also contain an embedded copy of FathomFox (see below)."

docs/installing.rst
```diff
@@ -6,6 +6,17 @@ Fathom consists of 3 parts. Here's how to install each one.
...
+
+    pip3 install fathom-web
+
+It's possible your Python package manager is called simply "pip" rather than "pip3". Give that a try if the above fails.
+
+You will also need to install `Node.js <https://nodejs.org/en/>`_ to use the most important commandline tools, like :doc:`fathom-train<commands/train>`.
```
erikrose, line 18: "to use many of the commandline tools." (fathom-train, -test, -histogram, -fox). It's not really necessary to say which, since they'll complain when you run them.

docs/rules.rst
```diff
@@ -143,3 +143,78 @@ Getting Ready to Train
 ======================
 
 Once you've written a few scoring rules, it's time for :doc:`training`. This computes optimal weighting coefficients and biases, which let the ruleset compute accurate confidences of its decisions. Be sure to write at least 2 rules before attempting to train; when there's only one, the trainer has nothing to balance.
+
```
erikrose, line 146: I think, if we're going to talk about trainees, it should be up under "Starting Your Ruleset", where I kind of half-bakedly already do it. Maybe, with enough conceptual explanation of the trainee data structure, a link to the code will suffice rather than another inevitably drifting excerpt?

docs/training.rst
```diff
@@ -52,13 +52,13 @@ Viewing the TensorBoard graphs with ``tensorboard --logdir runs/`` will quickly
...
 
-   Incidentally, it's not the end of the world if some scores go slightly outside [0, 1]. Limited tests have gotten away with values up to about 10 without measurable harm to training speed or accuracy. However, when feature values differ in magnitude by a factor of 1000, annoying oscillations dominate early iterations. Stick to [0, 1] for a trouble-free experience.
+   Incidentally, it's not the end of the world if some scores from rules go slightly outside [0, 1]. Limited tests have gotten away with values up to about 10 without measurable harm to training speed or accuracy. However, when feature values differ in magnitude by a factor of 1000, annoying oscillations dominate early iterations. Stick to [0, 1] for a trouble-free experience.
```
erikrose, line 61: Or "some rules' scores" for flow

docs/training.rst
```diff
@@ -109,7 +109,7 @@ Identifying Misrecognized Elements
...
...are insufficient to identify them. To start the process of using the Evaluator, open an instance of Firefox with FathomFox and your ruleset loaded (:doc:`fathom-fox<commands/fox>` makes this simple), and then open all of the samples you want to diagnose as separate tabs. Next, open the Evaluator page using the FathomFox's browser action button. The Evaluator page is initially quite empty: You have a dropdown to select your trainee ID and a button to run the Evaluator. When you click Evaluate, any pages with misrecognized nodes will show up in red; click those to see which element was wrongly selected. Unfortunately, you need to manually show the dev tools and switch to the Fathom panel once you get to the page in question; there aren’t yet web extension APIs to do it automatically. Once you do, you’ll see a quick and dirty representation of the “bad” element: a new label called “BAD [the trainee ID]”. Be sure to delete this if you choose to re-save the page for some reason. Also note that the BAD label is created only when the bad cell is clicked, for speed; if you navigate to the bad page manually, the label won’t be there, or there might be an old label from a previous iteration.
```
erikrose, line 112: ~~start the process of using~~  -> "use"

erikrose: We might as well just inline the whole relevant syntax for fathom-fox: `fathom-fox -r ruleset.js`.

fathom/fnode.mjs
```diff
@@ -8,8 +8,8 @@ import {getDefault, setDefault, sigmoid} from './utilsForFrontend';
  */
 export class Fnode {
     /**
-     * @arg element The DOM element I describe
-     * @arg ruleset The ruleset which created me
+     * @arg element The DOM element described by the fnode.
```
erikrose, line 11: the → this

fathom/fnode.mjs
```diff
@@ -41,7 +41,7 @@ export class Fnode {
     }
 
     /**
-     * Return whether the given type is one of the ones attached to the node.
+     * Return whether the given type is one of the ones attached to the fnode.
```
erikrose, line 44: I really was talking about the HTML node from here down. Maybe we say "HTML node I wrap" or something?

fathom_fox/addon/pages/evaluate.html
```diff
@@ -133,14 +133,14 @@ <h1>Evaluator</h1>
...
+      Evaluate your ruleset, for the selected trainee, against the samples you’ve loaded as tabs in this window.
     </p>
     <div class="uiGrid">
       <div class="label">
-        <label for="ruleset">Ruleset:</label>
+        <label for="trainee">Trainee ID:</label>
       </div>
       <div>
-        <select id="ruleset"></select>
+        <select id="trainee"></select>
```
erikrose, line 143: This probably breaks initRulesetMenu(), which finds the menu by ID.

docs/training.rst
erikrose: Yes, this is kind of a big block of text, especially to keep your place in while trying to follow its instructions. How about an ordered list?

docs/rules.rst
danielhertenstein: I agree about pointing toward the example code. I think adding an additional highlight of the `trainees` object in "Starting Your Ruleset" should be enough. We're telling them to use the example as a starting point, so they _should_ wind up reading that code (which explains the `trainees` object) anyway.

Random thought: Perhaps a `fathom-init` tool could be useful for setting up a directory structure and the skeleton of the trainees file.

fathom/fnode.mjs
danielhertenstein: Whoops, I was just overeager in my replacements.

danielhertenstein: Actually, isn't the type attached to the fnode and not the node?

fathom_fox/addon/pages/evaluate.html
danielhertenstein: Indeed it does.

erikrose:
Yes, but I'm not trying to emphasize the data structure here; I'm trying to teach concepts. Conceptually, the node has a type, no matter what it's directly hanging off of.

docs/rules.rst
erikrose: I have kicked around a "fathom-init" as well. :-) I'd take one. A few things get made implicitly, like `vectors` and `runs`, but it wouldn't be a bad thing to have `samples` , `samples/{validation,testing,training,unused}`, `rubrics.md`, and the example `rulesets.js` laid down. And we could automate `.gitattributes`. Hmm, I'm liking this more and more. FIts my usual evolutionary narrative:

1. Do it by hand.
2. Document it.
3. Script it.
4. Make it unnecessary altogether. :-)

danielhertenstein:
Okay, that works for me!

docs/training.rst
```diff
@@ -109,7 +109,17 @@ Identifying Misrecognized Elements
...
+FathomFox's Evaluator can point out misrecognized elements, in case the tag exerpts emitted by the trainer are insufficient to identify them. To use the Evaluator:
+
+#. Open an instance of Firefox with FathomFox and your ruleset loaded (``fathom-fox -r rulesets.js`` makes this simple).
+#. Open all of the samples you want to diagnose as separate tabs.
+#. Open the Evaluator page using FathomFox's browser action button.
+#. In the Trainee ID dropdown, select the trainee ID you want to diagnose.
```
erikrose, line 117: We can just say "trainee" rather than "trainee ID".

fathom_fox/addon/pages/evaluate.html
```diff
@@ -133,14 +133,14 @@ <h1>Evaluator</h1>
       No rulesets found. Please download a source checkout of FathomFox, and define one in rulesets.js.
     </p>
     <p>
-      Evaluate your ruleset based on the samples you’ve loaded as tabs in this window.
+      Evaluate your ruleset, for the selected trainee, against the samples you’ve loaded as tabs in this window.
     </p>
     <div class="uiGrid">
       <div class="label">
-        <label for="ruleset">Ruleset:</label>
+        <label for="trainee">Trainee ID:</label>
```
erikrose, line 140: How about just "Trainee"?

docs/training.rst
danielhertenstein: I think so.

PR #248, opened 2020-07-07
Combine Fathom commands into a single one. Close #233.

### LINKED ISSUE #233: Make fathom CLI tools 1 command
Have a top-level `fathom` command with subcommands under it for `train`, `test`, etc. This would make our tools more discoverable, because there'd be a top-level `fathom --help` or `fathom help` to call.

Example: https://github.com/mozilla/dxr/blob/master/dxr/cli/__init__.py

erikrose:
Shorten the first line of each subcommand's docstring, because click insists on truncating these in `fathom --help`. Tweak the rest of the docstrings so they don't become awkwardly redundant.

### DIFF
--- a/cli/fathom_web/commands/__init__.py
+++ b/cli/fathom_web/commands/__init__.py
@@ -1,0 +1,27 @@
+from .extract import extract
+from .fox import fox
+from .histogram import histogram
+from .label import label
+from .list import list
+from .pick import pick
+from .serve import serve
+from .test import test
+from .train import train
+
+from click import group
+
+
+@group()
+def fathom():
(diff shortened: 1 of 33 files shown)


erikrose:
This does make tab completion not work as well (e.g. `fathom-hist[TAB]`). Bad enough to give up?

erikrose:
I was going to leave those as they were, because they're accurate as of when they were written. The links still work. It's akin to how we still refer to fathom-trainees in old version notes even though that no longer exists. Do you feel strongly about it?

danielhertenstein:
I do not.

erikrose:
Thanks!

PR #250, opened 2020-07-15
Cast ints to floats in MCC computation to avoid potential integer overflows

danielhertenstein:
Casting objects feels bad to me. I think it's because I assume it points to some other problem, but this also seems a simple way to avoid the potential problem.

### DIFF
--- a/cli/fathom_web/accuracy.py
+++ b/cli/fathom_web/accuracy.py
@@ -196,7 +196,7 @@ def pretty_accuracy(description, accuracy, number_of_samples, false_positives, f
         precision = true_positives / (true_positives + false_positives)
     # Recall is the same as the true positive rate:
     recall = 1 - false_negative_rate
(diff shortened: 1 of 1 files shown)


erikrose:
[Ints have unlimited precision in Python 3.](https://docs.python.org/3/library/stdtypes.html?highlight=integer#numeric-types-int-float-complex) Take a look:

```python
>>> 1000000**4
1000000000000000000000000
>>> type(100000**4)
<class 'int'>
```

Even if each cell of the confusion matrix had 1 million samples, we'd still be okay. Or am I missing something?

danielhertenstein:
Ah, it looks like `accuracy_per_tag()` returns `numpy.int32`s which do not have infinite precision. Since those values are only used as `int`s, perhaps we're better casting there?

erikrose:
Ah, yes, that sounds like a good plan. Fewer surprises for callers. :-)

And sorry I didn't see your context from Matrix before I first commented.

Issue #256, opened 2020-07-21
Duplicate feature vectors for samples appearing in vector file


danielhertenstein:
While working on fathom-form-autofill, I have noticed feature vectors for samples appearing twice in the vector file created by the vectorizer. I have not confirmed if duplication occurs every time. The sample that is duplicated is not the same every time, though I have not confirmed if it occurs for only a subset of the samples or any of them.

For the next couple weeks, I will apply a bandage solution to fix the problem in a post-processing step.

After the initial fathom-form-autofill effort is completed, I will find the source of the problem and correct it.

PR #257, opened 2020-07-21
Fixing duplicated vectors in a post-processing step.

danielhertenstein:
This is a bandage solution for #256. Any duplicated feature vectors get removed at the same time that the file hashes are added to the header after vectorization. Any qualms with this being an in-place operation? As I stated in the linked issue, I will solve the problem more appropriately after the initial form autofill effort is completed.

### DIFF
--- a/fathom_fox/addon/vector.js
+++ b/fathom_fox/addon/vector.js
@@ -137,6 +137,16 @@ class Vectorizer extends PageVisitor {
     }
 
     async processAtEndOfRun() {
+        // Remove potential duplicated feature vectors from the vector file.
+        // 7/24/2020: This is a bandage solution to the problem of duplicated
+        // feature vectors. I (Daniel) will fix the source of the problem in
+        // ~2-4 weeks. If you are reading this message after that time has
+        // passed, please tell me to fix the problem and get rid of this code.
+        let filenamesSeen = new Set();
+        this.vectors = this.vectors.filter(item => {
+            return filenamesSeen.has(item['filename']) ? false : filenamesSeen.add(item['filename']);
+        });
+
         function compareByKey(key) {
             function cmp(a, b) {
                 const keyA = key(a);
--- a/fathom_fox/addon/vector.js
+++ b/fathom_fox/addon/vector.js
@@ -142,10 +142,8 @@ class Vectorizer extends PageVisitor {
         // feature vectors. I (Daniel) will fix the source of the problem in
         // ~2-4 weeks. If you are reading this message after that time has
         // passed, please tell me to fix the problem and get rid of this code.
-        let filenamesSeen = new Set();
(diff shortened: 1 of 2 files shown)


fathom_fox/addon/vector.js
```diff
@@ -137,6 +137,16 @@ class Vectorizer extends PageVisitor {
     }
 
     async processAtEndOfRun() {
+        // Remove potential duplicated feature vectors from the vector file.
+        // 7/24/2020: This is a bandage solution to the problem of duplicated
+        // feature vectors. I (Daniel) will fix the source of the problem in
+        // ~2-4 weeks. If you are reading this message after that time has
+        // passed, please tell me to fix the problem and get rid of this code.
+        let filenamesSeen = new Set();
```
erikrose, line 145: Could be `const` since we're never rebinding it. I tend to prefer `const` where possible because it makes things just a teense easier to reason about.

fathom_fox/addon/vector.js
```diff
@@ -137,6 +137,16 @@ class Vectorizer extends PageVisitor {
...
+        this.vectors = this.vectors.filter(item => {
+            return filenamesSeen.has(item['filename']) ? false : filenamesSeen.add(item['filename']);
```
erikrose, line 147: Looks good, but we can tersen like this:

```js
this.vectors = this.vectors.filter(item => filenamesSeen.has(item['filename']) ? false : filenamesSeen.add(item['filename']));
```

I like your clever use of the set, returned from `add`, as a true value. Even empty sets are true in JS. :-D

erikrose: It occurs to me that de-duping by filename will spuriously remove things if a filename occurs multiple times in the case of nested-folder corpora. But we don't do that much. And we have a ticket open to put in longer pathnames.

fathom_fox/addon/vector.js
danielhertenstein: Oh whoops, yes that should be `const`.

fathom_fox/addon/vector.js
danielhertenstein: Do you care about line length here? The form you suggest goes to 132 characters. And you are right this would cause problems with reoccurring filename. We could solve that by looking at the contents of `nodes` in addition to the filename. I had originally just thrown the entire array into a set, but forgot that the timing portion for each file will be different.

erikrose: I'll take comprehensibility and tokenwise brevity every time over short lines.

Since apparently the visitor is really visiting the duplicate pages more than one (the timings being different), I suppose the nodes could be different as well, them being occasionally nondeterministic. Filename should be fine for a quick patch.

erikrose: If you want, you can parenthesize to wrap the lines.

PR #262, opened 2020-11-06
Write docs on how to read metrics and review model PRs

erikrose:
I based these on the notes you took during our meeting. All clear? Anything left out? Thanks!

### DIFF
--- a/docs/maintaining.rst
+++ b/docs/maintaining.rst
@@ -1,0 +1,17 @@
+===========
+Maintaining
+===========
+
+A successful production model will need to be improved from time to time.
+
+Reviewing a Ruleset Change
+==========================
+
+The points to consider when reviewing a model change:
+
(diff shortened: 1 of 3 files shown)


docs/maintaining.rst
```diff
@@ -0,0 +1,17 @@
...
+
+  Ideally you can collect several samples representative of the problem you're trying to solve and distribute them across the training/validation/test sets. If you can find only one, you'll have to settle for putting it in training so the coefficients can be informed by it.
+* Make sure the "before" and "after" metrics, with commandline flags, are in the commit message to justify the change.
+* Review ruleset code changes as in a normal code review, for correctness and comprehensibility.
+* Check that any added samples have been extracted and render properly in Firefox. Use :doc:`fathom serve<commands/serve>` to make sure cross-origin policies (which are picky for ``file://`` URLs) aren't preventing the loading of subresources.
```
mnoorenberghe, line 17: Nit: I think it could be useful having separate sections for ruleset changes vs. corpus additions. In the latter I would put this point plus checking of the sample naming (assuming a naming convention is common).

mnoorenberghe (approved):
Thanks for writing this up Erik!

docs/maintaining.rst
erikrose: I just gave it a try, but the bullets don't divide up neatly. A big chunk of the first and the entire second are applicable to both sections. So I'll add the naming-convention point and go with that. Thank you for having a look!

erikrose: Actually, I ended up with a "If you added samples to the corpus, do these as well:" section. That worked.

PR #267, opened 2021-01-06
CircleCI

erikrose:
Get CircleCI builds going, now that Travis CI free service is going away.

PR #276, opened 2021-03-16
Update geckodriver to fix JS tests on CircleCI. (Props to glenda!)

erikrose:


erikrose:
Oops, see #267 instead.

PR #277, opened 2021-03-16
 Closes #178 Added evaluation of accuracy to determine cut off threshold

### LINKED ISSUE #178: Optimize confidence cutoffs
In addition to optimizing coefficients, have fathom-train pick an optimal value for the confidence cutoff. For example, if we get more training samples correct by requiring >70% than >80%, pick 70%, emit it to the user, and judge our accuracies based on it. (Then use that same threshold for the validation set, of course.) Currently, the trainer assumes a hard-coded 50% cutoff. Our binary cross-entropy loss function largely pushes the samples to the far end of the confidence space, so we haven't bothered being hypervigilant about this in the past, but we might be able to eek out another percent or two by doing this one simple thing.

gleonard-m:
Closes #178  Evaluates threshold values, displays result 

- Determines the optimal cut off using the mid point between predicted sigmoid values as possible cutoffs.  The cut off is rounded to 2 decimal places.

- Removed `--confidence-threshold, -t` from fathom train options

**Output sample**
`Optimal cutoff: 0.31`


cli/fathom_web/commands/train.py
```diff
@@ -54,12 +54,34 @@ def learn(learning_rate, iterations, x, y, confidence_threshold, num_prunes, val
...
+    print(f'All listed thresholds have the same accuracy value of {max_accuracy}')
+
     # Horizontal axis is what confidence. Vertical is how many samples were that confidence.
     writer.add_histogram('confidence', confidences(model, x), t)
     writer.close()
     return model
 
 
+def find_optimal_threshold(y, y_pred, num_prunes, threshold_incr=0.05):
+    max_range = int(1.0 / threshold_incr) + 1
```
erikrose, line 70: It seems to me (and I could be wrong) that there's a better way to go at this than trying a set of static positions in [0, 1]. Accuracy changes only when we move the threshold to the other side of a datapoint. So all we really need to do is slide the accuracy past each of the datapoints in turn, measuring accuracy at each. And, for best expected results in an infinite universe of discourse, we'd put our final threshold halfway between the 2 datapoints it divides. This removes the need for a threshold and guarantees optimal accuracy, even if datapoints are closer together than `threshold_incr`. What do you think?

cli/fathom_web/commands/train.py
```diff
@@ -54,12 +54,34 @@ def learn(learning_rate, iterations, x, y, confidence_threshold, num_prunes, val
     if stopped_early:
         print(f'Stopping early at iteration {t}, just before validation error rose.')
 
+    print(f'Successfully trained model using threshold {confidence_threshold}.')
+    optimal_thresholds, max_accuracy = find_optimal_threshold(y, y_pred, num_prunes, threshold_incr=0.05)
+
+    print(f'Suggested threshold(s) are: {optimal_thresholds}.')
+    print(f'All listed thresholds have the same accuracy value of {max_accuracy}')
```
erikrose, line 61: We won't need to say this, since the giant accuracy chart already displays accuracy.

cli/fathom_web/commands/train.py
```diff
@@ -54,12 +54,34 @@ def learn(learning_rate, iterations, x, y, confidence_threshold, num_prunes, val
     if stopped_early:
         print(f'Stopping early at iteration {t}, just before validation error rose.')
 
+    print(f'Successfully trained model using threshold {confidence_threshold}.')
```
erikrose, line 57: Do we need to say this? Is it actionable for the user? I've tried to minimize the output of `fathom train` (though I've failed for the "Building Fathomfox...done" bits).

cli/fathom_web/commands/train.py
gleonard-m: For each datapoint?  You're not concerned with the performance?  How many datapoints are usually expected?

cli/fathom_web/commands/train.py
gleonard-m: That chart will display the accuracy for the threshold at 0.5?

erikrose:
So far we’ve had thousands-ish, and maybe we can vectorize. I’d much rather have a better model than faster training.

cli/fathom_web/commands/train.py
erikrose: (Replying from my phone didn't maintain the thread structure, so I'm moving this in-thread by hand.)

So far we’ve had thousands-ish, and maybe we can vectorize. I’d much rather have a better model than faster training. Though, if it takes seconds, we should just go with what you've got. I expect the exhaustive search will be O(n^2), given that we'd have to evaluate every datapoint for every threshold position, unless we get clever. I don't think we could straight binary-search, but there's probably some pruning to be had. Regardless, we shouldn't get fancy unless we need to. :-)

cli/fathom_web/commands/train.py
erikrose: Well, `print_per_tag_report` will use whatever confidence threshold you pass in, and we should pass in our optimized threshold. Similar, `pretty_accuracy` (which prints the table of numbers with the confusion matrix and so on) [takes the output of `accuracy_per_tag`](https://github.com/mozilla/fathom/blob/489e78f2fd857ad64b259933b0566b7e4373f3c3/cli/fathom_web/commands/train.py#L269-L275), which itself takes the threshold. So, if we pass our optimized threshold to that as well, it should be reflected in all our output.

cli/fathom_web/commands/train.py
```diff
@@ -46,18 +51,53 @@ def learn(learning_rate, iterations, x, y, confidence_threshold, num_prunes, val
...
+
+
+def get_possible_cutoffs(y_pred):
+    """Using y_pred gets the sigmoid values, rounds the values and then gets the unique list.
+    This will reduce the number of cutoffs to be evaluated."""
+    with torch.no_grad():
+        flattened = y_pred.sigmoid().numpy().flatten()
+        cutoffs = np.sort(flattened)
+        if len(cutoffs) > 1:
+            cutoffs = np.unique([round((current + next) / 2, CUTOFF_DECIMAL_PLACES) for current, next in pairwise(cutoffs)])
```
erikrose, line 76: Very clear. I like the use of pairwise().

cli/fathom_web/commands/train.py
```diff
@@ -46,18 +51,53 @@ def learn(learning_rate, iterations, x, y, confidence_threshold, num_prunes, val
...
+        return cutoffs
```
erikrose, line 77: Does the right thing happen if there's only 1 (unique) datapoint? I think it could go either way. Let's say every datapoint had a value of 0.5, and some were positives (T) and some negatives (F). accuracy_per_tag() calls it a positive if the datapoint ≥ the cutoff. So this set would work out in our favor:
`0.5T 0.5T 0.5T`

And this set would be all wrong:
`0.5F 0.5F 0.5F`

So really, we always need to consider at least 2 cutoffs: in the case of a single unique value, some iota to the left and some iota to the right.

I'll leave this part as an exercise to the reader, but do you think we always need to consider a possible cutoff position an iota to the left of the min value and to the right of the max? It might be possible to construct a pathological dataset that needs those in the case of multiple unique values, and, if it is useful there, it would save a special case.

cli/fathom_web/test/test_train.py
```diff
@@ -32,3 +33,13 @@ def test_auto_vectorization_smoke(tmp_path):
     )
     assert result.exit_code == 0
     assert (tmp_path / 'training_vectors.json').exists()
+
+
+def test_find_optimal_cutoff():
```
erikrose, line 38: I guess with all those corner cases that popped up above, we'll need some more tests! :-)

erikrose (changes requested):
Thanks, Glenda! This is much simpler. But a bunch of corner-case trolls wandered out from under the bridge, and I need your help beating them back. :-o It is dangerous to go alone. Here, take these: ⚔️.

cli/fathom_web/commands/train.py
```diff
@@ -46,40 +51,53 @@ def learn(learning_rate, iterations, x, y, confidence_threshold, num_prunes, num
...
     if stopped_early:
         print(f'Stopping early at iteration {t}, just before validation error rose.')
 
-    print(f'Successfully trained model using threshold {confidence_threshold}.')
-    optimal_thresholds, max_accuracy = find_optimal_threshold(y, y_pred, num_prunes, threshold_incr=0.05)
-
-    print(f'Suggested threshold(s) are: {optimal_thresholds}.')
-    print(f'All listed thresholds have the same accuracy value of {max_accuracy}')
-
+    optimal_cutoff = find_optimal_cutoff(y, y_pred, num_prunes)
```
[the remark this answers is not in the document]
gleonard-m: yep, missed that somehow.

cli/fathom_web/commands/train.py
```diff
@@ -46,18 +51,53 @@ def learn(learning_rate, iterations, x, y, confidence_threshold, num_prunes, val
...
+
+def find_optimal_cutoff(y, y_pred, num_prunes):
+    """Evaluates possible cutoff values using accuracy as the metric.
+    If more than 1 cutoff gives the highest accuracy the midpoint
+    between the min and max cutoff is used."""
+    max_accuracy = 0
+    optimal_cutoffs = []
+
+    possible_cutoffs = get_possible_cutoffs(y_pred)
+    for _, test_cutoff in enumerate(possible_cutoffs):
```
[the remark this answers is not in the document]
gleonard-m: yep

cli/fathom_web/commands/train.py
```diff
@@ -46,18 +51,53 @@ def learn(learning_rate, iterations, x, y, confidence_threshold, num_prunes, val
...
+        if accuracy == max_accuracy:
+            optimal_cutoffs.append(test_cutoff)
+        if accuracy > max_accuracy:
+            optimal_cutoffs = [test_cutoff]
+            max_accuracy = accuracy
+
+    # Should always have at least 1 cutoff unless something has really gone sideways.
+    optimal_cutoff = optimal_cutoffs[0]
+    if len(optimal_cutoffs) > 1:
+        optimal_cutoff = round((optimal_cutoffs[0] + optimal_cutoffs[len(optimal_cutoffs) - 1]) / 2, CUTOFF_DECIMAL_PLACES)
```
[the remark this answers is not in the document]
gleonard-m: I've decided to go with calculating the mean of the possible cutoffs and selecting the value from the possible cutoffs that is closest to the mean.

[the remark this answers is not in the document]
erikrose: I assume when you say "possible", you mean "optimal". (Correct me if necessary.) Sounds fine. Since it's slightly more complicated to implement than just picking the middle item of the array, be sure to leave a comment explaining why you're doing it that way. It doesn't have to prove anything; if you're just following a hunch, say why you have the hunch and leave it at that. Thanks! :-D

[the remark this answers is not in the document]
gleonard-m: Yes "possible" should read "optimal".  Will do wrt code comments.

cli/fathom_web/commands/train.py
gleonard-m: Determining the optimal cutoff using a single (or 2 or 3) possible cutoff isn't adequate.  I'll look into a method to expand the list of possible cutoffs.

WRT always needing to consider a possible cutoff to the left on the min or right of the max:  

The only time adding another possible threshold higher than the calculated max possible threshold would have an impact on accuracy is if the optimal threshold is the highest threshold from the possibles. e.g [0.3, 0.4, 0.5] Adding 0.55 to the possibles list does not add any benefit if 0.4 is the optimal value.  If 0.5 is the optimal value then adding 0.55 to the possibles list may result in an increase in accuracy.

However the value of 0.5 is the midpoint of the sigmoid values for the 2 highest y_pred values.  Assuming that y is labeled a positive for the highest prediction value, moving the cutoff more to the right will not have an impact unless it goes beyond the sigmoid value of the highest y_pred.  That will result in all negatives giving a worse accuracy that the midpoint cutoff.  Does this make sense?  am I over simplifying?

I tested using a variety of y_pred values from [-2:2) inc by 0.01.  I also used [-2:1) incr by 0.01 with [2] appended to provide a large gap in the values at the high end. I could not find a dataset that showed improvement by adding a possible cutoff to the right of the max possible cutoff.

gleonard-m: To solve the problem of not having an adequate number of possible cutoff values I want to propose:

If there are less than a predetermined number of values (say 5),  for each possible cutoff pull 10 values from a normal distribution, truncated to [0,1] with mu set to the existing possible cutoff value and sigma low (current testing uses 0.1)  These values are then run through the same rounding and uniqueness processing to reduce the list.  This gives additional possible cutoffs that are close to the data driven cutoffs.

cli/fathom_web/commands/train.py
```diff
@@ -76,6 +78,28 @@ def possible_cutoffs(y_pred):
...
+    the midpoint of the list will result in the same optimal cutoff value being selected.
+    In the case where there is a larger, skewed list of values using the mean softens the
+    effect of the excessively skewed list.
+    """
+    cutoffs_mean = statistics.mean(cutoffs)
+    pos = bisect_left(cutoffs, cutoffs_mean)
+
+    if pos == 0:
+        return cutoffs[0]
+    if pos == len(cutoffs):
```
erikrose, line 106: len(cutoffs) - 1

erikrose (changes requested):
Functionality looks great, Glenda! Thank you! I just had some things to say about the tests, which I think are largely things left over from when the stuff they're testing was more complicated. I haven't had time to get through all the tests tonight, but I'll continue at test_find_optimal_cutoff_single_cutoff_with_highest_accuracy() when I sit down again.

cli/fathom_web/test/test_train.py
```diff
@@ -32,3 +33,146 @@ def test_auto_vectorization_smoke(tmp_path):
...
+
+    # Get the accuracy for each possible cutoff
+    # Note this is a different method of tracking the best cutoff than in find_optimal_cutoff.
+    cutoff_accuracy = {}
+    for possible in possibles:
+        accuracy, _, _ = accuracy_per_tag(y, y_pred, possible, num_prunes=0)
+        cutoff_accuracy[possible] = accuracy
+
+    # Get the key/value that has the highest accuracy.
+    expected_cutoff = max(cutoff_accuracy.items(), key=operator.itemgetter(1))[0]
```
erikrose, line 98: Could factor out `max(cutoff_accuracy.items(), key=operator.itemgetter(1))`.

cli/fathom_web/test/test_train.py
```diff
@@ -32,3 +33,146 @@ def test_auto_vectorization_smoke(tmp_path):
...
+    max_accuracy = max(cutoff_accuracy.items(), key=operator.itemgetter(1))[1]
+
+    # Verify that there is only 1 cutoff with this max accuacy.
+    accuracy_occurrences = 0
+    for accuracy in cutoff_accuracy.values():
```
erikrose, line 103: `sum(1 for accuracy in cutoff_accuracy.values() if max_accuracy == accuracy)` if you like. Never mind if not.

cli/fathom_web/test/test_train.py
```diff
@@ -32,3 +33,146 @@ def test_auto_vectorization_smoke(tmp_path):
...
+    y_pred = tensor([-2, -1.5, -1, -0.5, 0, 0.5, 1, 1.25, 2, 2.5])
+    y = tensor([0., 1., 0., 1., 0., 1., 0., 1., 0., 1.])
+
+    # Determining the expected_cutoff
+    possibles = possible_cutoffs(y_pred)  # this list will contain the optimal cutoff
+    assert len(possibles) == 9
+
+    # get the accuracy for each possible cutoff
+    # Note this is a different method of tracking the best cutoff than in find_optimal_cutoff.
+    cutoff_accuracy = {}
```
erikrose, line 128: I'd rather not have this loop (and maybe the next one too) repeated across 2 tests; it's a nontrivial amount of code, and factoring it out would make it both quicker to understand the test suite and less prone to bugs and pointless accumulation of differences.

cli/fathom_web/test/test_train.py
```diff
@@ -32,3 +33,146 @@ def test_auto_vectorization_smoke(tmp_path):
...
+    for possible in possibles:
+        accuracy, _, _ = accuracy_per_tag(y, y_pred, possible, num_prunes=0)
+        cutoff_accuracy[possible] = accuracy
+
+    # Get the cutoffs with the max accuracy, there should be more than 1
+    # (to verify the case where more than one cutoff has the best accuracy).
+    max_accuracy = max(cutoff_accuracy.items(), key=operator.itemgetter(1))[1]
+    best_cutoffs = []
+    for cutoff, accuracy in cutoff_accuracy.items():
```
erikrose, line 137: This is a textbook place for a list comprehension:
```
best_cutoffs = [cutoff for cutoff, accuracy in cutoff_accuracy.items() if max_accuracy == accuracy]
```

cli/fathom_web/test/test_train.py
```diff
@@ -32,3 +33,146 @@ def test_auto_vectorization_smoke(tmp_path):
...
+        if max_accuracy == accuracy:
+            best_cutoffs.append(cutoff)
+
+    # Verifying that there are more than 1 cutoff with the best accuracy (the basis for this test case)
```
erikrose, line 141: Great comment; making the intent explicit was very helpful to me

erikrose (changes requested):
Alright, that's all I've got! Thanks again, Glenda!

cli/fathom_web/test/test_train.py
```diff
@@ -32,3 +33,146 @@ def test_auto_vectorization_smoke(tmp_path):
...
+    assert possibles == expected
+
+    # Reduces to single cutoff
+    y_pred = tensor([1.2512, 1.2516])
+    expected = [0.78]
+    possibles = possible_cutoffs(y_pred)
+    assert len(possibles) == 1
+    assert all([a == b for a, b in zip(possibles, expected)])
+
+    y_pred = tensor([1.2512, 1.2516, 1.255])
```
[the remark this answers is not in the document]
gleonard-m: It does in a way.  A single y_pred results in 1 possible cutoff, 2 y_pred values also result in 1 possible cutoff due to using the midpoint.  Using 3 y_pred values will give 2 possibles that should then be reduced to a single value due to rounding.

cli/fathom_web/test/test_train.py
```diff
@@ -32,3 +33,146 @@ def test_auto_vectorization_smoke(tmp_path):
...
+    expected = [0.78]
+    possibles = possible_cutoffs(y_pred)
+    assert len(possibles) == 1
+    assert all([a == b for a, b in zip(possibles, expected)])
+
+    y_pred = tensor([1.2512, 1.2516, 1.2518, 1.2525, 1.2560])
```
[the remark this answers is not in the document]
gleonard-m: No, the case is covered by the previous test.  Will remove.

cli/fathom_web/test/test_train.py
```diff
@@ -32,3 +33,146 @@ def test_auto_vectorization_smoke(tmp_path):
...
+
+    # Partial reduction in number of cutoffs
+    y_pred = tensor([-2.1605, -0.5696, 0.4886, 0.8633, -1.3479,
+                     -0.5813, -0.5696, 0.5696, -0.5950, -0.5696])
+    expected = [0.15, 0.28, 0.36, 0.49, 0.63, 0.67]
+    possibles = possible_cutoffs(y_pred)
+    assert len(possibles) == 6
+    assert all([a == b for a, b in zip(possibles, expected)])
+
+    # No reduction in number of cutoffs
```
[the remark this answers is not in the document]
gleonard-m: The midpoints are used as possible cutoff, 4 y_pred values = 3 cutoffs.

cli/fathom_web/test/test_train.py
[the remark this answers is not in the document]
erikrose: Aha, makes sense! Sorry for missing that.

erikrose (approved):
Looks good! Go ahead and merge it in! I'll leave it up to you if you want to clean up any of the commit history. All I care about is having a nice merge-commit message. You get a gold star for sticking with this unexpected marathon!

